PR #21, opened 2022-10-18
make sure we only slice a valid size

khyberspache:


privateducky (approved):

privateducky:
@khyberspache I think your PR from today scooped this change up better

Issue #82, opened 2022-11-15
update default configuration to prod


makk94:
Right now, on `prelude configure`, the default  hq still points at our dev env - probably should change that to prod hq before we launch

PR #85, opened 2022-11-15
Default url to prod

mireaulf:
Closes https://github.com/preludeorg/libraries/issues/82

### DIFF
--- a/python/cli/prelude_cli/templates/template.cs
+++ b/python/cli/prelude_cli/templates/template.cs
@@ -5,7 +5,7 @@ CREATED: $CREATED
 */
 using System;
 
-class TTP {
+class Question {
     static void Test() {
         Console.WriteLine("Run test");
         Environment.Exit(100);
--- a/python/sdk/prelude_sdk/models/account.py
+++ b/python/sdk/prelude_sdk/models/account.py
@@ -29,13 +29,13 @@ def verify_credentials(func):
 
 class Account:
 
-    def __init__(self, profile='default', hq='https://detect.dev.prelude.org', keychain_location=os.path.join(Path.home(), '.prelude', 'keychain.ini')):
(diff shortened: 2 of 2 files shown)


python/cli/prelude_cli/templates/template.cs
```diff
@@ -5,7 +5,7 @@
 */
 using System;
 
-class TTP {
+class Question {
```
privateducky, line 8: can you make sure all templates in cli/ui have the same class name?

privateducky (approved):

python/cli/prelude_cli/templates/template.cs
privateducky: probably... SecurityTest ?

python/cli/prelude_cli/templates/template.cs
```diff
@@ -5,7 +5,7 @@
 */
 using System;
 
-class TTP {
+class SecurityTest {
```
kenrick, line 8: Ok will need to update Build as well to keep these in sync

mireaulf: PR is already opened

mireaulf: https://github.com/preludeorg/build/pull/129

kenrick: Awesome!

makk94:
can we have a single "main" document for CLI intro/guide - either the readme here or the official docs site - and then just redirect the other doc to the main one? Less stuff to remember to update/ keep in sync

privateducky (approved):

mireaulf:
> can we have a single "main" document for CLI intro/guide - either the readme here or the official docs site - and then just redirect the other doc to the main one? Less stuff to remember to update/ keep in sync

That would be great and readme.com allows us to update the docs through the api, do I hear that this is what you would like to work on next?

Issue #95, opened 2022-11-16
create-account stomps local account credentials


khyberspache:
We should add confirmation and/or not stomp the default keychain.ini credentials when making a create-account call

PR #104, opened 2022-11-17
Prompt to overwrite profile with iam create-account

mireaulf:
Closes https://github.com/preludeorg/libraries/issues/95

### DIFF
--- a/python/cli/prelude_cli/views/iam.py
+++ b/python/cli/prelude_cli/views/iam.py
@@ -1,5 +1,5 @@
 import click
-from prelude_cli.views.shared import handle_api_error
+from prelude_cli.views.shared import handle_api_error, prompt_overwrite_credentials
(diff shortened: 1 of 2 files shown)


privateducky (changes requested):
```
@click.confirmation_option(prompt='Are you sure?')
```

mireaulf:
I didn't went with that because it's too generic. If a user types in `prelude iam create-account` and you ask him `Are you sure?`. He will most likely say yes every time as it doesn't really tell what is going to happen if he accepts.

privateducky:
> I didn't went with that because it's too generic. If a user types in `prelude iam create-account` and you ask him `Are you sure?`. He will most likely say yes every time as it doesn't really tell what is going to happen if he accepts.

you can change the prompt text?

PR #140, opened 2022-12-01
Add windows for Swift probe

mireaulf:
Closes https://github.com/preludeorg/roadmap/issues/5

### DIFF
--- a/swift/probe/Sources/moonlight.swift
+++ b/swift/probe/Sources/moonlight.swift
@@ -81,8 +81,7 @@ struct System {
     func executable(url: URL) -> Optional<URL> {
         var path = URL(fileURLWithPath: url.path)
         #if os(Windows)
-        let range = url.lastPathComponent.range(of: ".tmp")!
-        path = URL(string: "\(url.lastPathComponent[..<range.lowerBound]).exe", relativeTo: path)!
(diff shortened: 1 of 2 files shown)


swift/probe/Sources/moonlight.swift
```diff
@@ -79,16 +78,28 @@ struct System {
         task.waitUntilExit()
         return task.terminationStatus
     }
-    func executable(url: URL) {
+    func executable(url: URL) -> Optional<URL> {
+        var path = URL(fileURLWithPath: url.path)
+        #if os(Windows)
+        let range = url.lastPathComponent.range(of: ".tmp")!
+        path = URL(string: "\(url.lastPathComponent[..<range.lowerBound]).exe", relativeTo: path)!
```
makk94, line 85: since you already have this PR, I'll just add here instead of opening my own :)
I came up with the same basic answer, but if it helps:

can replace the .tmp extension with: 
```
path = url.deletingPathExtension().appendingPathExtension("exe")
```

and a nit, but can use `FileManager.default.moveItem` instead of `copyItem`

mireaulf: moveItem will fail if the file already exists on disk, which happened during testing

mireaulf:
The size of the output binary is 29k but it requires the Swift runtime to be installed

PR #152, opened 2022-12-03
test: setup harness using vitest

kenrick:
### Description
This PR sets up the test harness for the javascript sdk using vitest.

Reopened because #146 was closed


kenrick:
@privateducky because we squash and merge I believe it safer to merge from latest pr. Knowing that those changes will be included in this PR.

privateducky:
> @privateducky because we squash and merge I believe it safer to merge from latest pr. Knowing that those changes will be included in this PR.

Let's merge together on Monday.

Originally, I was thinking we'd put these tests somewhere not in this repo, but now that I see it, I like where you're going. I think I'd back your directories out one, so each language has an optional 'test' directory. currently, all directories have SDK, CLI, probe (all optional).

PR #183, opened 2023-01-10
update bash demo & add powershell

makk94:
updated bash demo:
- updated wording & coloring per https://github.com/preludeorg/roadmap/issues/46#issuecomment-1375786732
- better error handling (print response code & exit early)
- sending max of (testresult, cleanresult) when saving results

bash demo tested on darwin m1
ps demo tested on windows 11

### DIFF
--- a/shell/probe/demo.sh
+++ b/shell/probe/demo.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+GREEN='\033[1;32m'
+RED='\033[1;31m'
+NC='\033[0m' # No Color
+
 PRELUDE_API="https://api.preludesecurity.com"
 PRELUDE_TOKEN=$1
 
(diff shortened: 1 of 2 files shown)


shell/probe/demo.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+GREEN='\033[1;32m'
```
privateducky, line 3: how's this green look on your ultra green terminal?

privateducky (approved):
Great work.
Let's get some testing on this on alternative machines and record compatibility. 
Specifically: platform (and version), architecture, shell used (for PowerShell, record version).

shell/probe/demo.sh
makk94: lol now some of the green is bold, can you tell?

Screen Shot 2023-01-10 at 5 30 30 AM

PR #187, opened 2023-01-10
linux arm64 vs aarch64

makk94:


shell/probe/install.sh
```diff
@@ -5,7 +5,12 @@ PRELUDE_ACCOUNT_ID=""
 PRELUDE_ACCOUNT_SECRET=""
 PROBE_NAME="nocturnal"
 ENDPOINT_TOKEN=""
-DOS="$(uname | awk '{print tolower($0)}')-$(uname -m)"
+
+arch=$(uname -m)
+if [ $arch = "aarch64" ];then
```
privateducky, line 10: Do we need something similar in nocturnal?

privateducky: Curious if we should just send this in as the DOS and give Service (backbone -> probe) a lookup table.. otherwise we may be playing catch up

privateducky (changes requested):
Preferably in backbone

shell/probe/install.sh
makk94: mm yeah, added nocturnal too

we could push the conversion to our backend.. i have no strong opinions

privateducky: Let's do it. I bet we uncover a few more things like this from alternative systems

makk94: not sure if I like it better, but 
https://github.com/preludeorg/backbone/pull/28 and
https://github.com/preludeorg/service/pull/497

PR #200, opened 2023-01-13
Add support for unexpected lookup value

mireaulf:


khyberspache:
Does this catch the 126 code too?

privateducky:
> Does this catch the 126 code too?

this will catch it by default.. but we should line up all our expected codes in this enum to return values we already know @mireaulf

mireaulf:
@privateducky Added the issue https://github.com/preludeorg/libraries/issues/201

PR #202, opened 2023-01-13
override api with env variable

makk94:


privateducky (approved):

shell/probe/demo.sh
```diff
@@ -4,7 +4,7 @@ GREEN='\033[1;32m'
 RED='\033[1;31m'
 NC='\033[0m' # No Color
 
-PRELUDE_API="https://api.preludesecurity.com"
+PRELUDE_API=${PRELUDE_API:="https://api.preludesecurity.com"}
```
privateducky, line 7: @makk94 can you just triple-check this syntax will work on all 'normal' shell environments? linux/macOS, bash, zsh, sh.

makk94: tested on my mac (arm, zsh) and my linux vm (x86_64, bash, and sh)

privateducky: thanks!

PR #217, opened 2023-01-19
 handling windows defender in demo & raindrop

makk94:
PR wraps our VST invocation in a try/catch to catch when the file fails to run b/c Defender prevents it

`ApplicationFailedException` is thrown when Defender prevents the test from running - I'm not sure if this exception is also thrown for other reasons though. There's no good documentation on what throws this. If we want to be certain, we could do string parsing of the exception message, but there's always a risk to error message changes IMO

Example of output with this code:
Screen Shot 2023-01-18 at 9 44 03 PM

### DIFF
--- a/shell/probe/demo.ps1
+++ b/shell/probe/demo.ps1
@@ -39,7 +39,7 @@ function DownloadTest {
 function ExecuteTest {
     try {
         $p = Start-Process -FilePath $TempFile -Wait -NoNewWindow -PassThru
-        if ($p.ExitCode -eq 100 ) {
+        if ($p.ExitCode -in 100,9,17,18,105,127 ) {
             Write-Host -ForegroundColor Green "`r`n[$($symbols.CHECKMARK)] Result: control test passed"
         } else {
             Write-Host -ForegroundColor Red "`r`n[!] Result: control test failed"
--- a/shell/probe/demo.ps1
+++ b/shell/probe/demo.ps1
@@ -38,38 +38,40 @@ function DownloadTest {
 
 function ExecuteTest {
     try {
-        & $TempFile
-        if ($LASTEXITCODE -eq 100 ) {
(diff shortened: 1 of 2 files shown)


shell/probe/demo.ps1
```diff
@@ -37,19 +37,39 @@ function DownloadTest {
 }
 
 function ExecuteTest {
-    & $TempFile
-    if ($LASTEXITCODE -eq 100 ) {
+    try {
+        & $TempFile
```
mireaulf, line 41: Use the following so that you can get the exit staus correctly
```
$p = Start-Process -FilePath <path> -ArgumentList <args> -Wait -NoNewWindow -PassThru
$p.ExitCode
```

makk94: Does that provide different results for you? It doesn't for me

privateducky: It may or may not, depending on the happy or unhappy paths. on the latter, you will see differences. 
but regardless, we need start process to run tests in a child process.

makk94: done

mireaulf (approved):

shell/probe/demo.ps1
```diff
@@ -131,7 +153,7 @@ Write-Host "
 ###########################################################################################################
 "
 
-if ($Status -eq 100 ) {
+if ($Status -in 100,127 ) {
```
privateducky, line 156: more codes then this exist... we'd be better off putting all codes in here and printing one of four conditions:
- passed
- failed
- error
- defense caught

privateducky: or at minimum; combining passed and defense code codes into "worked"

mireaulf (approved):

privateducky (approved):

PR #222, opened 2023-01-19
Set env system wide

mireaulf:


shell/probe/install.ps1
```diff
@@ -73,7 +73,7 @@ if(Test-Path -path $probePath -PathType Leaf) {
 LogMessage "Determining OS"
 $dos = "windows-" + $Env:PROCESSOR_ARCHITECTURE
 $token=RegisterEndpoint
-$Env:PRELUDE_TOKEN = $token
+[System.Environment]::SetEnvironmentVariable("PRELUDE_TOKEN", $token)
```
makk94, line 76: Think this still defaults to current process

Either set scope or I've been trying out `setx`

PR #224, opened 2023-01-19
specify TLS protocol

makk94:
solves this error: `The request was aborted: Could not create SSL/TLS secure channel.`

By default powershell uses TLS 1.0 and looks like we require TLS 1.2


privateducky (approved):
should we require

PR #225, opened 2023-01-19
Environment variable consistency

mireaulf:
Getting an env var in powershell can be done multiple ways
image

### DIFF
--- a/shell/probe/install.ps1
+++ b/shell/probe/install.ps1
@@ -82,7 +82,8 @@ if(Test-Path -path $probePath -PathType Leaf) {
 LogMessage "Determining OS"
 $dos = "windows-" + $Env:PROCESSOR_ARCHITECTURE
 $token=RegisterEndpoint
-[System.Environment]::SetEnvironmentVariable("PRELUDE_TOKEN", $token, "User")
+[Environment]::SetEnvironmentVariable("PRELUDE_TOKEN", $token, "User")
+[Environment]::SetEnvironmentVariable("PRELUDE_API", $PRELUDE_API, "Process")
 DownloadProbe $token $dos $probePath
 StartTask $token $parentDir $probePath
 Write-Host "[=] Detect setup complete"
--- a/shell/probe/install.ps1
+++ b/shell/probe/install.ps1
@@ -10,6 +10,8 @@ param(
   [String]$endpointId=$env:computername
 )
 
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+
 function FromEnv { param ([string]$envVar, [string]$default)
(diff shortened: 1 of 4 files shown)


shell/probe/install.ps1
```diff
@@ -10,8 +10,15 @@ param(
   [String]$endpointId=$env:computername
 )
 
+function FromEnv { param ([string]$envVar, [string]$default)
```
makk94, line 13: nit: move params to next line just to be consistent with other function headers

makk94: also a small nit: be consistent across the files wrt if FromEnv function comes before or after the TLS  setting

makk94 (approved):

shell/probe/install.ps1
```diff
@@ -11,7 +11,15 @@ param(
 )
 
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
-$PRELUDE_API=if ($Env:PRELUDE_API) { $Env:PRELUDE_API } else { "https://api.preludesecurity.com" }
+
+function FromEnv { param ([string]$envVar, [string]$default)
+    $envVal = [Environment]::GetEnvironmentVariable($envVar, "Process")
+    if($envVal) {return $envVal}
```
makk94, line 17: if it's only a process-scoped variable, should you also set the user-scoped variable of the same name so the installed probe will work?

makk94: or could update the cmd we're passing the scheduled task to set the process-scoped env variable before starting raindrop

mireaulf: The install script sets the PRELUDE_TOKEN at User-level assuming that the script is not modified by a user, I don't see how the PRELUDE_TOKEN at Process-level would override User-level

makk94: was thinking about the PRELUDE_API variable - just so that the probe you install is talking to the right environment

mireaulf: Ah good catch!

shell/probe/install.ps1
```diff
@@ -74,7 +82,8 @@ if(Test-Path -path $probePath -PathType Leaf) {
 LogMessage "Determining OS"
 $dos = "windows-" + $Env:PROCESSOR_ARCHITECTURE
 $token=RegisterEndpoint
-[System.Environment]::SetEnvironmentVariable("PRELUDE_TOKEN", $token, "User")
+[Environment]::SetEnvironmentVariable("PRELUDE_TOKEN", $token, "User")
+[Environment]::SetEnvironmentVariable("PRELUDE_API", $PRELUDE_API, "Process")
```
makk94, line 86: the ScheduledTask will start a different process I think, so you need to add this line to the scheduled task action I believe

privateducky (approved):

PR #228, opened 2023-01-20
updates probes (return test result always, have demo handle exit codes better)

makk94:
Bringing other probes up-to-date with recent windows changes:
* return test outcome only, not `max(test,clean)`
* update demo to account for other potential exit codes


privateducky (approved):
I assume Moonlight is already good?

Issue #221, opened 2023-01-19
Remove install from demo.sh


kenrick:
Relates to https://github.com/preludeorg/build/issues/376

PR #240, opened 2023-01-21
export endpoint token between demo and install to reuse same token

makk94:
re: https://github.com/preludeorg/libraries/issues/221

will update powershell in a separate PR

### DIFF
--- a/shell/probe/demo.sh
+++ b/shell/probe/demo.sh
@@ -5,7 +5,7 @@ RED='\033[1;31m'
 NC='\033[0m' # No Color
 
 PRELUDE_API=${PRELUDE_API:="https://api.preludesecurity.com"}
-ENDPOINT_TOKEN=$1
+PRELUDE_TOKEN=$1
 
 sys=$(uname -s)-$(uname -m)
 id="b74ad239-2ddd-4b1e-b608-8397a43c7c54"
@@ -17,7 +17,7 @@ function check_relevance {
 
 function download_test {
     temp=$(mktemp)
-    location=$(curl -sfSL -w %{url_effective} -o $temp -H "token:${ENDPOINT_TOKEN}" -H "dos:${dos}" -H "id:${id}" $PRELUDE_API)
+    location=$(curl -sfSL -w %{url_effective} -o $temp -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "id:${id}" $PRELUDE_API)
     test=$(echo $location | grep -o '[0-9a-f]\{8\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{12\}' | head -n 1)
     if [ -z "$test" ];then
         echo -e "${RED}[!] Failed to download test${NC}"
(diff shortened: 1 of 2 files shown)


privateducky (commented):
I don't understand the change set here; most of it is renaming the env var, where the only applicable line appears to be 65 (export VARIABLE) in the demo.sh.

makk94:
> I don't understand the change set here; most of it is renaming the env var, where the only applicable line appears to be 65 (export VARIABLE) in the demo.sh.

That's all correct. The renaming is just for consistency/ I think it's clearer. The exporting is the functional change - it goes with the changes in install.sh.   If user chooses to install a probe after running the demo script, the demo script exports ENDPOINT_TOKEN. meanwhile install.sh checks if ENDPOINT_TOKEN is set as an env variable, and if so, will reuse that token instead of registering a new endpoint

privateducky:
Why ENDPOINT TOKEN instead of PRELUDE though? Referring to the same thing by two names across our code base can get confusing. (I initially chose PRELUDE, so all our env vars, currently 3, could be spotted easily).

makk94:
I just think ENDPOINT_TOKEN is more explicit - I've had multiple people just at Prelude ping me that "probes aren't working" b/c they are using an admin token instead of an endpoint token. what about PRELUDE_ENDPOINT_TOKEN across the board? It matches PRELUDE_ACCOUNT_ID and PRELUDE_ACCOUNT_SECRET better

 (it also nags at me that we use "secret" when referring to accounts and "token" for endpoints, not sure if you feel the same ha)

privateducky:
We can definitely look at renaming these vars but let's separate that from this PR.

Renaming the vars is a larger change: docs, code, probes, handling existing deployments, etc, so while I'm all for it, we should isolate why someone would get confused first (poor docs, poor error messaging, etc).

makk94:
Moved demo.sh back to PRELUDE_TOKEN (though that wasn't a public variable/argument so it shouldn't have been a problem...) 

Also moved ENDPOINT_TOKEN in install.sh to an arg to stay consistent with other args there

privateducky (changes requested):
I don't want to invent new env var names in one spot, so let's use our existing names:
- PRELUDE_TOKEN
- PRELUDE_API
- PRELUDE_CA

For account ID/token pairs, we currently do not leverage these in env vars, so we can invent something for our install/demo scripts, assuming they are all consistent. PRELUDE_ACCOUNT_ID / PRELUDE_ACCOUNT_SECRET are fine by me, although the latter is likely odd if the user is required to send in a 'token' as a header.

makk94:
> I don't want to invent new env var names in one spot, so let's use our existing names:
> 
> * PRELUDE_TOKEN
> * PRELUDE_API
> * PRELUDE_CA
> 
> For account ID/token pairs, we currently do not leverage these in env vars, so we can invent something for our install/demo scripts, assuming they are all consistent. PRELUDE_ACCOUNT_ID / PRELUDE_ACCOUNT_SECRET are fine by me, although the latter is likely odd if the user is required to send in a 'token' as a header.

I'm not adding any new env vars in this PR. Are you requesting that the install.sh argument descriptor names match the env vars? Ie. you want the help menu to look like:
```
mahinakaholokula@Mahinas-MacBook-Pro libraries % ./shell/probe/install.sh -h

Usage: install.sh [-h] -n PROBE_NAME

  -h                          Shows Usage
  -n PROBE_NAME               Probe Name; Default: nocturnal
  -t PRELUDE_TOKEN            Endpoint Token; Use PRELUDE_ACCOUNT_ID and PRELUDE_ACCOUNT_SECRET to register a new endpoint
  -a PRELUDE_ACCOUNT_ID       Prelude Account Id; asdf
  -s PRELUDE_ACCOUNT_SECRET   Prelude Account Secret; 
  -e ENDPOINT_ID              Endpoint Id; Default: Mahinas-MacBook-Pro.local
```

makk94:
@privateducky Originally was pushing the endpoint_token through so that a re-run of the demo script would work (endpoint token would still be active). With the trouble were having and looking at https://github.com/preludeorg/range/issues/13 - I'm just going to switch to re-registering the endpoint in install.sh. The demo script can pass through the endpoint id and install.sh will re-register with same endpoint name/ get a new token back

privateducky:
@makk94 exactly; match the env var names with the names in the params.

Issue #239, opened 2023-01-21
Install scripts should support endpoint tags


privateducky:
Default tags to none (or [] or "").

PR #252, opened 2023-01-23
add endpoint tag option on install

makk94:
re: https://github.com/preludeorg/libraries/issues/239

Our backend currently only accepts a list, not a comma-separated list - if you're waiting for that update, it's coming, just lmk here!

### DIFF
--- a/shell/probe/install.ps1
+++ b/shell/probe/install.ps1
@@ -32,7 +32,7 @@ function LogMessage {
 
 function RegisterEndpoint {
     LogMessage "Provisioning Detect Endpoint Token..."
-    $data = @{"id"=$endpointId;"tags"=$endpointTags + ,"windows"} | ConvertTo-Json
+    $data = @{"id"=$endpointId;"tags"=$endpointTags} | ConvertTo-Json
     $response = Invoke-WebRequest -Method POST -Uri $PRELUDE_API/detect/endpoint -UseBasicParsing -Headers @{"account"=$preludeAccountId;"token"=$preludeAccountSecret} -ContentType "application/json" -Body $data
     if($response.StatusCode -ne 200) {
         LogError "Endpoint failed to register! $($response.StatusDescription)"
--- a/shell/probe/install.ps1
+++ b/shell/probe/install.ps1
@@ -7,7 +7,9 @@ param(
   [Parameter(HelpMessage="Probe name")]
   [String]$probeName="raindrop",
   [Parameter(HelpMessage="Endpoint id")]
-  [String]$endpointId=$env:computername
+  [String]$endpointId="",
+  [Parameter(HelpMessage="Endpoint tags")]
+  [String[]]$endpointTags=@()
 )
 
 function FromEnv { param ([string]$envVar, [string]$default)
@@ -30,7 +32,7 @@ function LogMessage {
 
 function RegisterEndpoint {
     LogMessage "Provisioning Detect Endpoint Token..."
-    $data = @{"id"=$endpointId;"tag"="windows"} | ConvertTo-Json
+    $data = @{"id"=$endpointId;"tags"=$endpointTags + ,"windows"} | ConvertTo-Json
     $response = Invoke-WebRequest -Method POST -Uri $PRELUDE_API/detect/endpoint -UseBasicParsing -Headers @{"account"=$preludeAccountId;"token"=$preludeAccountSecret} -ContentType "application/json" -Body $data
     if($response.StatusCode -ne 200) {
         LogError "Endpoint failed to register! $($response.StatusDescription)"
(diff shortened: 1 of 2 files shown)


shell/probe/install.sh
```diff
@@ -49,7 +54,9 @@ done
 register_new_endpoint() {
     echo "[+] Provisioning Detect Endpoint Token..."
     local _token_url="${PRELUDE_API}/detect/endpoint"
-    ENDPOINT_TOKEN=$(curl -sfS -X POST -H "account:${PRELUDE_ACCOUNT_ID}" -H "token:${PRELUDE_ACCOUNT_SECRET}" -H "Content-Type: application/json" -d "{\"id\":\"${ENDPOINT_ID}\",\"tag\":\"darwin\"}"  "${_token_url}")
+    local _tags="$(echo "[\"darwin$ENDPOINT_TAGS\"]" | sed 's/,/\",\"/g')"
```
mireaulf, line 57: Can we not hardcode darwin here, this function is used by all OSes

makk94: good catch, i was just blindly copy pasting

it feels kinda opinionated for us to add system tag at all for the user, but maybe doesnt matter too much

mireaulf (approved):

shell/probe/install.ps1
```diff
@@ -30,7 +32,7 @@ function LogMessage {
 
 function RegisterEndpoint {
     LogMessage "Provisioning Detect Endpoint Token..."
-    $data = @{"id"=$endpointId;"tag"="windows"} | ConvertTo-Json
+    $data = @{"id"=$endpointId;"tags"=$endpointTags + ,"windows"} | ConvertTo-Json
```
mireaulf, line 35: `,` on the wrong side of the double quote

makk94: it's actually on the correct side! it's dumb but when $endpointTags is empty, the comma there will make sure that powershell interprets "windows" as an array not a string. and when $endpointTags is not empty, then it's just adding to that list

mireaulf: 🤯

privateducky, line 35: default to empty; not + windows (I don't want to assume yet)

shell/probe/install.sh
```diff
@@ -5,8 +5,10 @@ PRELUDE_ACCOUNT_ID=""
 PRELUDE_ACCOUNT_SECRET=""
 PROBE_NAME="nocturnal"
 ENDPOINT_ID=$(hostname)
+ENDPOINT_TAGS=""
 ENDPOINT_TOKEN=""
-DOS="$(uname | awk '{print tolower($0)}')-$(uname -m)"
+sys=$(uname | awk '{print tolower($0)}')
```
privateducky, line 10: what's this change about?

privateducky (changes requested):
minor things

shell/probe/install.sh
makk94: original behavior of this script was to send in OS as a tag. install.sh was hardcoding "darwin", this line change was to account for 'linux' too. Since your other comment was to remove 'windows' as an automatic tag, I'll do the same here and this line change will be reverted

privateducky: ah, thanks!

mireaulf (approved):

shell/probe/install.sh
```diff
@@ -49,7 +54,12 @@ done
 register_new_endpoint() {
     echo "[+] Provisioning Detect Endpoint Token..."
     local _token_url="${PRELUDE_API}/detect/endpoint"
-    ENDPOINT_TOKEN=$(curl -sfS -X POST -H "account:${PRELUDE_ACCOUNT_ID}" -H "token:${PRELUDE_ACCOUNT_SECRET}" -H "Content-Type: application/json" -d "{\"id\":\"${ENDPOINT_ID}\",\"tag\":\"darwin\"}"  "${_token_url}")
+    if [[ $ENDPOINT_TAGS ]];
+    then
+      local _tags=",\"tags\":$(echo "[\"$ENDPOINT_TAGS\"]" | sed 's/,/\",\"/g')"
```
privateducky, line 59: this seems super dicey. can't we just take the string of comma-separated tags?

makk94: see original PR description: 

> Our backend currently only accepts a list, not a comma-separated list - if you're waiting for that update, it's coming, just lmk here!

privateducky: ;) yes please

privateducky: hmm; @makk94 are you going to add the BE support?

privateducky (approved):

privateducky:
@mireaulf have you tested this?

mireaulf:
> @mireaulf have you tested this?

Works

PR #266, opened 2023-01-24
move UNKNOWN to not be last

makk94:
we `create-user` with the lowest permission by default, which relies on our permissions list being ordered
`default=[p.name for p in Permission][-1]`'

moving UNKNOWN to the top of the list so that SERVICE is the "lowest" permission


mireaulf:
Not a huge fan of having to keep enum values ordered. Why not just changing the default of create-user to `Permission.SERVICE.name`?

makk94:
presumably in case we add a lower permission? i dont have strong preference between this and hardcoding SERVICE

privateducky (approved):

PR #277, opened 2023-01-26
only notify for prod release

packplusplus:


privateducky (changes requested):
isn't there a workflow file to remove?

PR #293, opened 2023-01-29
Update JS SDK to match Pythons

### LINKED ISSUE #292: JS SDK (master) needs to catch up to Python SDK
Several adjustments were made to the Python SDK on 1/28, which need to make their way into the JS SDK. 

These include:
- detect activity: swap from days param to start/finish
- build attachment routes are now GET:/build/tests/{id} and the return is not a list of attachments but a dict of attachments and mappings
- a new 'detect search' function was added
- 'build create-test' API route went from /build/tests to /build/tests/{id}

kenrick:
Fixes #292

### DIFF
--- a/javascript/sdk/index.html
+++ b/javascript/sdk/index.html
@@ -86,6 +86,8 @@
 
           service.setCredentials(credentials);
 
+          console.log("DETECT SEARCH", await service.detect.search("hi"));
+
           console.log("List Tests", await service.build.listTests());
 
           const testId = uuid.v4();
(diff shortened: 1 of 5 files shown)


javascript/sdk/index.html
```diff
@@ -86,6 +86,8 @@ <h1 class="text-xl py-2 ml-2">Testing SDK: (See Source)</h1>
 
           service.setCredentials(credentials);
 
+          console.log("DETECT SEARCH", await service.detect.search("hi"));
```
privateducky, line 89: if you want an actual good value back, select a CVE ID

javascript/sdk/lib/controllers/detect.ts
```diff
@@ -59,8 +60,12 @@ export default class DetectController {
   }
 
   /** Get report for an Account */
-  async describeActivity(days: number = 7, options: RequestOptions = {}) {
-    const searchParams = new URLSearchParams({ days: days.toString() });
+  async describeActivity(
+    start: string,
```
privateducky, line 64: both are optional, just so you know. defaults to start=30 days ago, finish = right now.

privateducky (approved):

Issue #282, opened 2023-01-26
Need to support argument for "writable directory" on install scripts + probes


khyberspache:
Some environments block tests from running out of `/tmp`

privateducky:
@khyberspache this one strikes me more as probe-related, right? the install scripts require write perms to the normal install locations but the probes are the ones using "unpredictable" temp dirs for dropping tests.. @makk94 can you check if this statement is true/false?

depending on that, we can adjust accordingly. either an env var, parameter, ability to find its own directory -- or a combo of these things, between the installer and probe.

makk94:
@privateducky wdyt of a PRELUDE_DIR env var that install script and probes all use, if set ?

privateducky:
> @privateducky wdyt of a PRELUDE_DIR env var that install script and probes all use, if set ?



> @privateducky wdyt of a PRELUDE_DIR env var that install script and probes all use, if set ?

This is probably the best option, as the situation is going to be rare enough. Both the install scripts and probes could honor that variable. If you want to code that up we can see how it looks/runs in practice.

makk94:
@khyberspache Did you have a specific customer/ situation in mind for this issue? Otherwise I think it's getting closed as WAI

privateducky:
> @khyberspache Did you have a specific customer/ situation in mind for this issue? Otherwise I think it's getting closed as WAI

Yes; we've adjusted the particular environment that motivated this ticket in the first place.

I'm debating whether to keep this ticket open on the issue board for the time being or kick ban it for a later date if we decide to support the "find" writable directory path.

privateducky:
@makk94 as I'm debating... wouldn't simply adjusting the installers to make the temp dir used by the probe writable for the probe user?

mireaulf:
@privateducky That would work

privateducky:
Updating scope on this ticket: this should be solved via the installers.

makk94:
talked to alex - we can't assume that we can create a writable/executable directory where the probe is. Our new nocturnal installer uses systemd and my understanding is that it can start the probe itself via EXECSTART but that doesn't guarantee that the probe is in a directory that is writable/executable. And we can't make a writable/executable directory b/c any filesystem mount flags (like noexec or readonly) might still exist (and we cant overwrite those)

@khyberspache did I get that right?

privateducky:
[...]

PR #298, opened 2023-01-29
add flag to choose download directory in install scripts; probes download vsts to cwd

makk94:
re: https://github.com/preludeorg/libraries/issues/282

### DIFF
--- a/go/probe/cmd/hades/hades.go
+++ b/go/probe/cmd/hades/hades.go
@@ -6,9 +6,8 @@ import (
 )
 
 func main() {
-	var ok bool
-	var hq string
-	if hq, ok = os.LookupEnv("PRELUDE_API"); !ok {
+	hq := os.Getenv("PRELUDE_API")
+	if hq == "" {
 		hq = "https://api.preludesecurity.com"
 	}
 	probe := hades.CreateProbe(os.Getenv("PRELUDE_TOKEN"), hq)
--- a/go/probe/internal/hades/hades.go
+++ b/go/probe/internal/hades/hades.go
@@ -39,7 +39,7 @@ func CreateProbe(token, hq string) *Probe {
         token:         token,
         hq:            strings.TrimSuffix(hq, "/"),
         dos:           strings.ToLower(fmt.Sprintf("%s-%s", runtime.GOOS, runtime.GOARCH)),
-        sleep:         3 * time.Second,
+        sleep:         14400 * time.Second,
         cwd:           wd,
         commandTimout: 2 * time.Second,
     }
--- a/go/probe/internal/hades/hades.go
+++ b/go/probe/internal/hades/hades.go
@@ -30,13 +30,10 @@ type Actions interface {
 }
 
 func CreateProbe(token, hq string) *Probe {
-    wd := os.Getenv("PRELUDE_DIR")
-    if wd == "" {
-        var err error
-        if wd, err = util.FindWorkingDirectory(); err != nil {
-            return nil
-        }
-    }
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil
+	}
     return &Probe{
         signals:       make(chan bool),
         token:         token,
@@ -76,7 +73,7 @@ func (p *Probe) runTask(data string) {
 }
 
 func (p *Probe) save(data []byte) (*os.File, error) {
-	f, err := os.CreateTemp(p.cwd, "detect")
+	f, err := os.CreateTemp(p.cwd, "detect-vst")
 	if err != nil {
 		return nil, err
 	}
--- a/go/probe/internal/hades/hades.go
+++ b/go/probe/internal/hades/hades.go
@@ -36,16 +36,16 @@ func CreateProbe(token, hq string) *Probe {
         if wd, err = util.FindWorkingDirectory(); err != nil {
             return nil
         }
-	}
-	return &Probe{
-		signals:       make(chan bool),
-		token:         token,
-		hq:            strings.TrimSuffix(hq, "/"),
-		dos:           strings.ToLower(fmt.Sprintf("%s-%s", runtime.GOOS, runtime.GOARCH)),
-		sleep:         3 * time.Second,
-		cwd:           wd,
-		commandTimout: 2 * time.Second,
-	}
+    }
+    return &Probe{
+        signals:       make(chan bool),
+        token:         token,
+        hq:            strings.TrimSuffix(hq, "/"),
+        dos:           strings.ToLower(fmt.Sprintf("%s-%s", runtime.GOOS, runtime.GOARCH)),
+        sleep:         3 * time.Second,
+        cwd:           wd,
+        commandTimout: 2 * time.Second,
+    }
 }
 
 func (p *Probe) Start() {
--- a/go/probe/internal/hades/hades.go
+++ b/go/probe/internal/hades/hades.go
@@ -30,9 +30,11 @@ type Actions interface {
 }
 
 func CreateProbe(token, hq string) *Probe {
-	wd, err := util.FindWorkingDirectory()
-	if err != nil {
-		return nil
+	if wd, ok = os.LookupEnv("PRELUDE_DIR"); !ok {
+        wd, err := util.FindWorkingDirectory()
+        if err != nil {
+            return nil
+        }
 	}
 	return &Probe{
 		signals:       make(chan bool),
--- a/go/probe/internal/hades/hades.go
+++ b/go/probe/internal/hades/hades.go
@@ -30,9 +30,10 @@ type Actions interface {
 }
 
 func CreateProbe(token, hq string) *Probe {
-	if wd, ok = os.LookupEnv("PRELUDE_DIR"); !ok {
-        wd, err := util.FindWorkingDirectory()
-        if err != nil {
+    wd := os.Getenv("PRELUDE_DIR")
+    if wd == "" {
+        var err error
+        if wd, err = util.FindWorkingDirectory(); err != nil {
             return nil
         }
 	}
@@ -41,7 +42,7 @@ func CreateProbe(token, hq string) *Probe {
 		token:         token,
 		hq:            strings.TrimSuffix(hq, "/"),
 		dos:           strings.ToLower(fmt.Sprintf("%s-%s", runtime.GOOS, runtime.GOARCH)),
-		sleep:         14400 * time.Second,
+		sleep:         3 * time.Second,
 		cwd:           wd,
 		commandTimout: 2 * time.Second,
 	}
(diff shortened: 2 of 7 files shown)


go/probe/cmd/hades/hades.go
```diff
@@ -6,9 +6,8 @@ import (
 )
 
 func main() {
-	var ok bool
-	var hq string
-	if hq, ok = os.LookupEnv("PRELUDE_API"); !ok {
+	hq := os.Getenv("PRELUDE_API")
```
privateducky, line 9: Pretty sure Go supports a fallback value when getting an env var

go/probe/internal/hades/hades.go
```diff
@@ -30,19 +30,22 @@ type Actions interface {
...
-		hq:            strings.TrimSuffix(hq, "/"),
-		dos:           strings.ToLower(fmt.Sprintf("%s-%s", runtime.GOOS, runtime.GOARCH)),
-		sleep:         14400 * time.Second,
-		cwd:           wd,
-		commandTimout: 2 * time.Second,
-	}
+    wd := os.Getenv("PRELUDE_DIR")
+    if wd == "" {
+        var err error
+        if wd, err = util.FindWorkingDirectory(); err != nil {
```
privateducky, line 36: This is different from our other probes.. I could see either the probe finding a writable dir or getting it as an env var but not both.

shell/probe/install.ps1
```diff
@@ -17,7 +17,7 @@ function FromEnv { param ([string]$envVar, [string]$default)
 
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 $PRELUDE_API = FromEnv "PRELUDE_API" "https://api.preludesecurity.com"
-
+$appDir = FromEnv "PRELUDE_DIR" (Join-Path ([System.Environment]::ExpandEnvironmentVariables("%LOCALAPPDATA%")) "prelude")
```
privateducky, line 20: Why join temp dir to a prelude dir? No other probe does this..

shell/probe/install.sh
```diff
@@ -110,7 +112,7 @@ EOF
 install_darwin() {
     local _running_user="${USER}"
     local _plist_path="/Users/${_running_user}/Library/LaunchAgents/com.preludesecurity.detect.plist"
-    local _app_dir="/Users/${_running_user}/.prelude/bin"
+    local _app_dir=${PRELUDE_DIR:="/Users/${_running_user}/.prelude/bin"}
```
privateducky, line 115: Why not use ~/tmp? This would work on Linux and Mac. Or just $mktemp

shell/probe/nocturnal.sh
```diff
@@ -7,7 +7,7 @@ dos=$(echo $sys | tr '[:upper:]' '[:lower:]')
 
 while :
 do
-    temp=$(mktemp)
+    if [ -z "$PRELUDE_DIR" ];then temp=$(mktemp); else temp="${PRELUDE_DIR}/$(uuidgen)"; fi
```
privateducky, line 10: Logic is too complex. I'm also not so sure how uuidgen will hold up on all distos.

privateducky (changes requested):

go/probe/cmd/hades/hades.go
makk94: lmk what that is - this is what I see as normal

go/probe/internal/hades/hades.go
makk94: I thought the point of this issue is to be able to overwrite the directory with the env var for the special cases that the directory we chose isn't writable/ executable? But us choosing should still be the default?

shell/probe/install.ps1
makk94: That was the default before - not sure if there was a reason for it. Removing now

shell/probe/install.sh
makk94: seems risky to install the probe in tmp, it might get autodeleted there

shell/probe/nocturnal.sh
makk94: what about `$RANDOM`  as the test name ? I'm under the assumption that it's a bad idea to choose a static filename (or filename prefix) like "prelude-vst" b/c that tips off EDRs - is that right?

go/probe/cmd/hades/hades.go
privateducky: Fallback, meaning a default value when the variable is not set. In most languages it is something like: get(var, fallback)

makk94: thx ha i know what a fallback is @privateducky 

im saying lmk the exact command in Go, because I didn't find one

privateducky: Might not exist then! Thought it did but not 100%

go/probe/internal/hades/hades.go
privateducky: Yep-- but this appears to not be using a temp dir but instead hunting for a writable directory. I'm reviewing on a phone so could be wrong (can't see the internals of that util func).

shell/probe/install.sh
privateducky: On a phone but.. is this for the permanent location of the probe binary or the initial location to download it to?

shell/probe/nocturnal.sh
privateducky: I'm not worried about the name per say, as if they squash our file and it's not malicious, we'll report as such. 

That said, if the probe was written to a writable directory via the installer, does the probe need to know about the location or can it just write tests to its cwd?

shell/probe/install.sh
makk94: this is the permanent location

privateducky: kk in that case, ignore me!

shell/probe/nocturnal.sh
makk94: I assume in most cases, users won't be setting PRELUDE_DIR, and for those, it'd be nice to keep writing to /tmp to clean up left over VSTs (from error exits)

If that's not the case/ we don't care about that, I can remove the if/else here and just have the VSTs always writing to cwd (and PRELUDE_DIR is removed as an env variable). It's up to the user to either run our installer with PRELUDE_DIR or download the probe themselves to a writable/executable location

privateducky: This would make sense to me. Pack the intelligence of where to put the probe in the installer and if someone skips using it, they'd need to assume that knowledge.

go/probe/internal/hades/hades.go
makk94: It does look like FindWorkingDirectory() walks from cwd to find a writable directory. No guarantee there's a writable directory to be found from cwd, but if there is, this function should find it

Obviously no other probe does this right now. Do we need all the probes to the same behavior, and is that behavior like this or like the others (default to /tmp)?

privateducky: Per our other thread; it makes most sense for the probe to write tests wherever it lives (or tmp) but if someone wants to specify a directory, the installer should have the logic to put the probe in a writable directory.

shell/probe/nocturnal.sh
makk94: okee. updated so that
- install scripts have PRELUDE_DIR env var option to override the default probe install directory we choose (`/opt` , or `~` generally)
- probes all download VSTs to cwd

tested:
- hades on darwin m1
- install.sh & nocturnal on darwin m1 and ubuntu
- install.ps1 and raindrop on windows 2019 (i think)

mireaulf (approved):

mireaulf:
Moonlight would also like that update

makk94:
> Moonlight would also like that update

Today we use a Foundation function (https://developer.apple.com/documentation/foundation/urlsession/1411511-downloadtask) that downloads the test and auto-cleans it up after we run the VST. That function chooses the location of the temporary file (on my mac, it's in /var) - since moonlight is super likely to only be run on macs i imagine, do you think we should trust that the location downloadTask chooses to write to is ok for our probe? The alternative is I add extra logic to copy the file from the temp location post download to cwd, and then manually delete the file after we run it

mireaulf:
> > Moonlight would also like that update
> 
> Today we use a Foundation function (https://developer.apple.com/documentation/foundation/urlsession/1411511-downloadtask) that downloads the test and auto-cleans it up after we run the VST. That function chooses the location of the temporary file (on my mac, it's in /var) - since moonlight is super likely to only be run on macs i imagine, do you think we should trust that the location downloadTask chooses to write to is ok for our probe? The alternative is I add extra logic to copy the file from the temp location post download to cwd, and then manually delete the file after we run it

We should still aim to support the same functionality for moonlight

makk94:
> We should still aim to support the same functionality for moonlight - i.e. provide an override with the PRELUDE_DIR param

The param override is only for the install scripts, not relevant to the probes. 

The probe logic for the other probes is to temporarily download the test to cwd and run it (under the assumption we won't write/execute permission issues if the test is in the same location at the probe itself) - so that'd be the only difference if we leave moonlight as is (probes would download and run the test from the os-chosen directory)

mireaulf (approved):

go/probe/internal/hades/hades.go
```diff
@@ -30,19 +30,19 @@ type Actions interface {
...
-		commandTimout: 2 * time.Second,
-	}
+    return &Probe{
+        signals:       make(chan bool),
+        token:         token,
+        hq:            strings.TrimSuffix(hq, "/"),
+        dos:           strings.ToLower(fmt.Sprintf("%s-%s", runtime.GOOS, runtime.GOARCH)),
+        sleep:         14400 * time.Second,
+        cwd:           wd,
+        commandTimout: 2 * time.Second,
```
privateducky, line 44: this PR needs to be rebased with the removal of the time out code

go/probe/internal/hades/hades.go
```diff
@@ -73,7 +73,7 @@ func (p *Probe) runTask(data string) {
 }
 
 func (p *Probe) save(data []byte) (*os.File, error) {
-	f, err := os.CreateTemp(p.cwd, "detect")
+	f, err := os.CreateTemp(p.cwd, "detect-vst")
```
privateducky, line 76: why?

privateducky (changes requested):
this topic makes me uncomfortable, because the changes required to do this should not be taken lightly. this touches every probe and every installation script, and opens up a number of things to consider that are new -- such as cleaning up files that are not in 'temp' locations, how to name files we put to disk, etc. 

I think our current stance can instead be: a host is incompatible if the probe is unable to write to the temp location.

PR #301, opened 2023-01-29
add request interceptor

kenrick:
This PR adds a request inceptor to allow users to change parts of the requests before its sent. Eg. Changing the _product based on which env the sdk it running in.

### DIFF
--- a/javascript/sdk/index.html
+++ b/javascript/sdk/index.html
@@ -88,6 +88,10 @@
         console.log("Using host", host);
         const service = new Prelude.Service({
           host,
+          requestInterceptor: (request) => {
+            request.headers.set("_product", "js-sdk-test");
+            return request;
+          },
         });
 
         const credentials = await service.iam.newAccount("user_handle");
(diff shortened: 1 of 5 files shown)


privateducky:
What's the goal?

kenrick:
A couple reasons; 
1. I want to remove this entire file in from build as just a wrapper around the sdk https://github.com/preludeorg/build/blob/master/packages/core/src/lib/api.ts. It's growing with every change we make to the sdk. 
2. One of the blockers to remove this file is we are overriding the _product header on every request to indicate if its a PWA or not. Current doing it here:https://github.com/preludeorg/build/blob/master/packages/core/src/lib/api.ts#L11. But now we can set it at the service level:

```typescript
 const service = new Prelude.Service({
   host,
   requestInterceptor: (request) => {
     request.headers.set("_product", isPWA() ? "js-sdk-pwa" : "js-sdk-web");
     return request;
   },
 });
```

privateducky (approved):
Makes sense!

PR #306, opened 2023-01-31
demo: make the get-test call prior to pressing enter

makk94:


privateducky (approved):
interesting; I like this a lot actually.
just confirm for me: on both scripts, we "find test" by getting the redirect URL, holding on to it, then when we "download test" we retrieve via the redirect.. right?

makk94:
> interesting; I like this a lot actually. just confirm for me: on both scripts, we "find test" by getting the redirect URL, holding on to it, then when we "download test" we retrieve via the redirect.. right?

That's correct. I tried downloading the binary to a variable for a while, couldn't figure it out. Also google tells me don't do that because you can't store null bytes in bash variables

privateducky:
> > interesting; I like this a lot actually. just confirm for me: on both scripts, we "find test" by getting the redirect URL, holding on to it, then when we "download test" we retrieve via the redirect.. right?
> 
> That's correct. I tried downloading the binary to a variable for a while, couldn't figure it out. Also google tells me don't do that because you can't store null bytes in bash variables

this is great. even better than what I had in mind - good work!

PR #313, opened 2023-02-03
Support filters on the query types

khyberspache:


python/cli/prelude_cli/views/detect.py
```diff
@@ -138,14 +138,25 @@ def rules(controller):
               help='retrieve a specific result view',
               default='logs', show_default=True,
               type=click.Choice(['logs', 'days', 'probes', 'insights'], case_sensitive=False))
+@click.option('--tests', help='a comma-separated list of test IDs to filter on', type=str)
+@click.option('--endpoint_ids', help='a comma-separated list of endpoint IDs to filter on', type=str)
```
privateducky, line 142: Think we can use "endpoints" instead of "endpoint_ids", similar to "tests"?

privateducky (commented):
1 tiny thing. Otherwise, awesome!

python/cli/prelude_cli/views/detect.py
khyberspache: Sure, I was just using the params name, but no biggie!

privateducky (approved):

PR #319, opened 2023-02-05
list fields should be plural

makk94:


privateducky (commented):
If we do this, we'll also need to update the API that handles these (my activity function).

makk94:
PRs out for that

privateducky (approved):

Issue #333, opened 2023-02-09
probe installed via install script doesnt restart on reboots [mac]


makk94:
repro'd on a couple different macs

PR #335, opened 2023-02-09
Install as LaunchDaemons

mireaulf:
Closes https://github.com/preludeorg/libraries/issues/333


mireaulf:
Install needs to run as sudo

`sudo -E -- ./install.sh ...`

PR #336, opened 2023-02-10
Install using the current users security principal

khyberspache:


privateducky:
@makk94 @mireaulf will wait on your thoughts

makk94:
Does this make the probe itself run w/ highest privileges too?

khyberspache:
> Does this make the probe itself run w/ highest privileges too?

No the probe runs with whatever privileges the current users token has

makk94:
I just ran this on a Windows 16 VM and after rebooting got:
`Task Scheduler did not launch task "\Prelude Detect"  because user "EC2AMAZ-9BFFCMC\Administrator" was not logged on when the launching conditions were met. User Action: Ensure user is logged on or change the task definition to allow launching when user is logged off.`

khyberspache:
> I just ran this on a Windows 16 VM and after rebooting got:
> `Task Scheduler did not launch task "\Prelude Detect" because user "EC2AMAZ-9BFFCMC\Administrator" was not logged on when the launching conditions were met. User Action: Ensure user is logged on or change the task definition to allow launching when user is logged off.`


We need to toggle the task definition to support ` change the task definition to allow launching when user is logged off.`

I'm getting more errors with the current pattern in an Active Directory environment

PR #341, opened 2023-02-13
Add get account route to iam

kenrick:
Closes: https://github.com/preludeorg/build/issues/570


privateducky (changes requested):
can you remove the 'list-users' and 'decide' routes?

kenrick:
@privateducky removed decide and list-users

Issue #351, opened 2023-02-15
CLI should support temporary user creation


privateducky:
Prelude accounts can now contain temporary users. There should be two changes to the CLI / API to support this from the client side:

1. List the expiry time per account in the 'iam account' command
2. Allow 'iam create-user' to pass in an optional expiry timestamp

mireaulf:
Depends on https://github.com/preludeorg/service/pull/635

privateducky:
https://github.com/preludeorg/libraries/commit/92db4b1872751c14267ef53fbc3a665866bc8ec1

PR #356, opened 2023-02-15
python-cli: support expiration

mireaulf:
Closes https://github.com/preludeorg/libraries/issues/351

### DIFF
--- a/python/cli/prelude_cli/views/iam.py
+++ b/python/cli/prelude_cli/views/iam.py
@@ -39,7 +39,7 @@ def describe_account(controller):
 @iam.command('create-user')
 @click.option('-p', '--permission', help='provide a permission level', default=[p.name for p in Permission][-1],
               type=click.Choice([p.name for p in Permission], case_sensitive=False), show_default=True)
-@click.option('-e', '--expires', help='provide an expiration timestamp; defaults to 1 year', type=int)
+@click.option('-e', '--expires', help='provide an expiration date (YYYY-MM-DD); defaults to 1 year', type=str)
 @click.argument('handle')
 @click.pass_obj
 @handle_api_error
--- a/python/cli/prelude_cli/views/iam.py
+++ b/python/cli/prelude_cli/views/iam.py
@@ -39,13 +39,14 @@ def describe_account(controller):
 @iam.command('create-user')
 @click.option('-p', '--permission', help='provide a permission level', default=[p.name for p in Permission][-1],
               type=click.Choice([p.name for p in Permission], case_sensitive=False), show_default=True)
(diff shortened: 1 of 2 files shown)


python/cli/prelude_cli/views/iam.py
```diff
@@ -39,7 +39,7 @@ def describe_account(controller):
 @iam.command('create-user')
 @click.option('-p', '--permission', help='provide a permission level', default=[p.name for p in Permission][-1],
               type=click.Choice([p.name for p in Permission], case_sensitive=False), show_default=True)
-@click.option('-e', '--expires', help='provide an expiration timestamp; defaults to 1 year', type=int)
+@click.option('-e', '--expires', help='provide an expiration date (YYYY-MM-DD); defaults to 1 year', type=str)
```
privateducky, line 42: match what we do on detect-activity

python/sdk/prelude_sdk/controllers/iam_controller.py
```diff
@@ -35,10 +35,10 @@ def get_account(self):
         raise Exception(res.text)
 
     @verify_credentials
-    def create_user(self, permission, handle):
+    def create_user(self, permission, handle, expires):
```
privateducky, line 38: this forces the user to set an expires value. we want that to be optional.

python/cli/prelude_cli/views/iam.py
```diff
@@ -31,20 +33,22 @@ def register_account(controller):
...
+@click.option('-d', '--days', help='days until expiration', default=365, type=int)
 @click.argument('handle')
 @click.pass_obj
 @handle_api_error
-def create_user(controller, permission, handle):
+def create_user(controller, permission, handle, days):
     """ Create a new user in your account """
-    token = controller.create_user(handle=handle, permission=Permission[permission.upper()].value)
-    click.secho(f'Created new [{permission}] user [{handle}]. Token: {token}', fg='green')
+    resp = controller.create_user(handle=handle, permission=Permission[permission.upper()].value, expires=datetime.utcnow() + timedelta(days=days))
```
privateducky, line 50: this is passing in a date time but no string formatting occurs. will this convert to JSON ok in the SDK function?

privateducky (changes requested):
a few comments

privateducky:
This code does not work.
1. The date time in the SDK is not JSON-convertable
2. The functionality just simply doesn't work. Why? Backbone required an updated query to ensure the insert succeeded. Expires, as a property, wasn't even in the SQL command at all. How did you test this?

I have fixed both areas and did a quick smoke test validation.

Issue #344, opened 2023-02-13
CLI: 'activity' endpoint should resolve test identifiers


privateducky:
For any 'detect activity' command, we should offer a boolean flag (--convert) that allows you to automatically convert any test ID to the plain-text name.

privateducky:
https://github.com/preludeorg/libraries/pull/366/files

PR #357, opened 2023-02-15
python-cli: show test names by default in activity

mireaulf:
Closes https://github.com/preludeorg/libraries/issues/344

### DIFF
--- a/python/cli/prelude_cli/views/detect.py
+++ b/python/cli/prelude_cli/views/detect.py
@@ -151,7 +151,7 @@ def add_recommendation(controller, title, description):
 @click.option('--endpoints', help='comma-separated list of endpoint IDs', type=str)
 @click.option('--dos', help='comma-separated list of DOS', type=str)
 @click.option('--statuses', help='comma-separated list of statuses', type=str)
-@click.option('-i', '--show-test-id', help='show test id instead of test names', is_flag=True, type=bool, default=False)
+@click.option('-e', '--expand', help='show test id instead of test names', is_flag=True, type=bool, default=False)
 @click.pass_obj
 @handle_api_error
 def describe_activity(controller, days, view, tests, tags, endpoints, dos, statuses, show_test_id):
--- a/python/cli/prelude_cli/views/detect.py
+++ b/python/cli/prelude_cli/views/detect.py
@@ -151,14 +151,16 @@ def add_recommendation(controller, title, description):
 @click.option('--endpoints', help='comma-separated list of endpoint IDs', type=str)
 @click.option('--dos', help='comma-separated list of DOS', type=str)
 @click.option('--statuses', help='comma-separated list of statuses', type=str)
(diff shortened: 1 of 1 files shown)


python/cli/prelude_cli/views/detect.py
```diff
@@ -172,8 +174,18 @@ def describe_activity(controller, days, view, tests, tags, endpoints, dos, statu
...
     report = Table()
+    build_tests = dict()
+
+    def get_test(test_id):
+        if show_test_id or not test_id:
+            return test_id
+        if not build_tests:
+            build = BuildController(account=controller.account)
+            build_tests.update({test['id']: test['name'] for test in build.list_tests()})
+        return build_tests[test_id]
```
makk94, line 185: print DELETED if test_id isn't in list_tests

makk94: at least that's what print somewhere else in our API

python/cli/prelude_cli/views/detect.py
```diff
@@ -151,14 +151,16 @@ def add_recommendation(controller, title, description):
 @click.option('--endpoints', help='comma-separated list of endpoint IDs', type=str)
 @click.option('--dos', help='comma-separated list of DOS', type=str)
 @click.option('--statuses', help='comma-separated list of statuses', type=str)
+@click.option('-i', '--show-test-id', help='show test id instead of test names', is_flag=True, type=bool, default=False)
```
privateducky, line 154: can we make this '-e' for '--expand' ?

python/cli/prelude_cli/views/detect.py
```diff
@@ -172,8 +174,18 @@ def describe_activity(controller, days, view, tests, tags, endpoints, dos, statu
 
     raw = controller.describe_activity(view=view, filters=filters)
     report = Table()
+    build_tests = dict()
+
+    def get_test(test_id):
+        if show_test_id or not test_id:
+            return test_id
+        if not build_tests:
```
privateducky, line 182: this is a lot of work; let's just do this in 2 lines: 1) get all tests 2) add to dict

mireaulf: What dict? Also I don't want to do that extra api call for views that don't need it which is the majority of them

mireaulf: I'm trying to not duplicate code here, I would have the same block of code for all views returning test ids

privateducky: hmm; my issue here isn't about normalizing the code, it's the fact ``build.list_tests()`` already returns a dict of all tests, so there is no need to move a test from that collection to a new ``build_tests`` dict on demand. just look the value up in the dictionary.

privateducky:
@mireaulf closing in favor of https://github.com/preludeorg/libraries/pull/366/files

note the difference. just a simple conversion lambda does the trick.

Issue #332, opened 2023-02-09
Windows doesn't clean up VST's from tmp until restart


gerbsec:
While Linux and MacOS have some functionality to clean up their respective TMP directories, windows will hoard onto its precious tmp directory until a restart occurs, which is often rare for server. This means that 100 tests later == 100 vsts in tmp just chilling there.

Random but also related, on linux and mac, using nocturnal, you can see not only the tmp vst's but also empty temp files. it seems like mktemp went through but a vst was not written to it.  same happens with raindrop.

privateducky:
Two distinct things here.

For the first, no Windows cleanup, we should adjust all probes to remove tests after executing them.

For the second, we need to investigate why this is.

makk94:
both of these just come down to "we don't clean up VSTs explicitly in probes"

(the second happens because we create the tempfile before we actually know if there's a test to write to it yet, so when there are no tests returned, you just get an empty file before the probe goes to sleep)

privateducky:
ah, makes sense. 
so adding the explicit test removal should do the trick.

privateducky:
I noticed raindrop is creating temp files before even retrieving a test. 
We should only create the temp file if a test is retrieved.

makk94:
I think it's the opposite. Nocturnal creates the temp files before retrieving the test. Raindrop doesn't I don't think

privateducky:
> I think it's the opposite. Nocturnal creates the temp files before retrieving the test. Raindrop doesn't I don't think

That's possible, but let's verify that. I see this at the very beginning of raindrop, which I believe will create a new file in Windows. This occurs before the request to Service:
```
function Run {
    Param([String]$Dat = "")

    $TempFile = [System.IO.Path]::GetTempFileName() | Rename-Item -NewName { [System.IO.Path]::ChangeExtension($_, "exe") } -PassThru
...
```

makk94:
Ah yeah youre right. In my choose-download-directory PR I had changed it to stop creating the temp files before getting the test. But that didn't actually get merged

I'll separate that logic out into a new PR

PR #361, opened 2023-02-17
Cleanup temp files

mireaulf:
Closes https://github.com/preludeorg/libraries/issues/332

### DIFF
--- a/shell/probe/nocturnal.sh
+++ b/shell/probe/nocturnal.sh
@@ -12,6 +12,7 @@ do
     test=$(echo $location | grep -o '[0-9a-f]\{8\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{12\}' | head -n 1)
 
     if [ -z "$test" ];then
+        rm -f $temp
         sleep 14400
     else
         ca=$(echo $location | sed -e 's|^[^/]*//||' -e 's|/.*$||')
@@ -27,4 +28,5 @@ do
             fi
         fi
     fi
+    rm -f $temp
 done
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -1,34 +1,36 @@
 function Run {
     Param([String]$Dat = "")
-
-    $TempFile = [System.IO.Path]::GetTempFileName() | Rename-Item -NewName { [System.IO.Path]::ChangeExtension($_, "exe") } -PassThru
-
-    $Headers = @{
-        'token' = $Token
-        'dos' = $Dos
-        'dat' = $Dat
-    }
     try {
-        $Response = Invoke-WebRequest -URI $Address -UseBasicParsing -Headers $Headers -MaximumRedirection 1 -OutFile $TempFile -PassThru
-    } catch {
-        $StatusCode = $_.Exception.Response.StatusCode.value__
-        Write-Output "ERROR: Failed to reach Prelude Service. " + $StatusCode
-        return
-    }
-    if ($CA -and $CA -ne $Response.BaseResponse.ResponseUri.Authority) {
-        return
-    }
-    $Test = $Response.BaseResponse.ResponseUri.AbsolutePath.Split("/")[-1].Split("_")[0]
-    if (-not $Test) {
-        Write-Output "INFO: Done running tests"
-        return
-    }
+        $TempFile = [System.IO.Path]::GetTempFileName() | Rename-Item -NewName { [System.IO.Path]::ChangeExtension($_, "exe") } -PassThru
 
-    $TestExit = Execute $TempFile
-    Start-Process -FilePath $TempFile -ArgumentList "clean" -Wait -NoNewWindow -PassThru
+        $Headers = @{
+            'token' = $Token
+            'dos' = $Dos
+            'dat' = $Dat
+        }
+        try {
+            $Response = Invoke-WebRequest -URI $Address -UseBasicParsing -Headers $Headers -MaximumRedirection 1 -OutFile $TempFile -PassThru
+        } catch {
+            $StatusCode = $_.Exception.Response.StatusCode.value__
+            Write-Output "ERROR: Failed to reach Prelude Service. " + $StatusCode
+            return
+        }
+        if ($CA -and $CA -ne $Response.BaseResponse.ResponseUri.Authority) {
+            return
+        }
+        $Test = $Response.BaseResponse.ResponseUri.AbsolutePath.Split("/")[-1].Split("_")[0]
+        if (-not $Test) {
+            Write-Output "INFO: Done running tests"
+            return
+        }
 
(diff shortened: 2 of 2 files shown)


shell/probe/nocturnal.sh
```diff
@@ -27,4 +28,5 @@ do
             fi
         fi
     fi
+    rm -f $temp
```
privateducky, line 31: why is this in two places? we should only require 1 clean up line.

shell/probe/raindrop.ps1
```diff
@@ -1,34 +1,36 @@
 function Run {
     Param([String]$Dat = "")
+    try {
```
privateducky, line 3: there is a lot of changes going into this file... hard to tell without a side by side, but is this entirely just a try/catch/finally change, to ensure the file is removed in a new finally block?

privateducky (changes requested):
one change, one question

shell/probe/nocturnal.sh
makk94: i assume the first is just a nicety so you don't have to stare at an empty temp file while the probe sleeps. An alternative is to not use mktemp so we don't create an empty file when there are no more tests left

shell/probe/raindrop.ps1
makk94: i believe so

mireaulf: That's right

shell/probe/nocturnal.sh
mireaulf: @makk94 is right here, I can remove the first one but the file will stay on disk until the probe comes back from sleep

privateducky: how does raindrop compare, functionality wise?

mireaulf: Raindrop uses try/finally

privateducky: I mean: how often are the temp files removed from the box?

mireaulf: Right after the test cleanup

makk94: if i'm reading raindrop right, we download and run all the tests and then we clean them all up right before sleeping, right? (we don't clean up between test runs)

in nocturnal, we clean up after every test run (and before sleeping)

privateducky: doesn't sound right... as I don't think we're storing a list of run tests in the loop, are we?

mireaulf: > if i'm reading raindrop right, we download and run all the tests and then we clean them all up right before sleeping, right? (we don't clean up between test runs)
> 
> in nocturnal, we clean up after every test run (and before sleeping)

Actually we only fetch one test and sleep...

mireaulf: > if i'm reading raindrop right, we download and run all the tests and then we clean them all up right before sleeping, right? (we don't clean up between test runs)
> 
> in nocturnal, we clean up after every test run (and before sleeping)

No. We fetch a single test and then sleep. Cleanup is done per test.

makk94: we recursively call Run tho

mireaulf: > we recursively call Run tho (so won't hit the finally until after all the recursions)
Updated raindrop to run after the temp file removal

privateducky: so what's the consensus?

mireaulf: Updated with a single `rm`

makk94: Another option - make this look more like demo.sh where the logic might be
1) get redirect url
2) if it doesn't have a uuid in it, sleep
3) else, follow the redirect, save the test to file, do the test, remove the file

then don't need to worry about removing a file before sleeping

privateducky (commented):
who has validated these probes, post changes?

shell/probe/nocturnal.sh
```diff
@@ -11,9 +11,9 @@ do
     location=$(curl -sL -w %{url_effective} -o $temp -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "dat:${dat}" $PRELUDE_API)
     test=$(echo $location | grep -o '[0-9a-f]\{8\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{12\}' | head -n 1)
 
+    ts = 0
     if [ -z "$test" ];then
-        rm -f $temp
-        sleep 14400
+        ts = 14400
```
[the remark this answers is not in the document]
makk94: oops. just updated it. I like my update, but lmk if you want it switched back to this

shell/probe/nocturnal.sh
makk94: closing because this thread is long

shell/probe/nocturnal.sh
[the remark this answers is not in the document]
mireaulf: That's cleaner

mireaulf:
I can't approve because it's my PR but ✅

privateducky:
> one change, one question

@mireaulf I was good w/this a few minutes ago.. but then a larger change set was just introduced.

mireaulf:
> > one change, one question
> 
> @mireaulf I was good w/this a few minutes ago.. but then a larger change set was just introduced.

Yes, saw that. It's been tested and I believe it's cleaner and closer to raindrop it terms of how it behaves

privateducky:
> > > one change, one question
> > 
> > 
> > @mireaulf I was good w/this a few minutes ago.. but then a larger change set was just introduced.
> 
> Yes, saw that. It's been tested and I believe it's cleaner and closer to raindrop it terms of how it behaves

This changes the behavior too much, which makes me uncomfortable. For example, it now hard codes /tmp, vs using the system's mktemp location. I'd prefer the previous version.. if you feel strongly that this is better, we will need to do a few rounds of review and testing as a team before this could be considered.

makk94:
mktemp first checks TMPDIR env variable, and defaults back to /tmp  - this PR does the same
(so on my mac, I ran this and it's still downloading to /var, not /tmp)

if that's not convincing, I'll revert back. just lmk!

privateducky:
> mktemp first checks TMPDIR env variable, and defaults back to /tmp - this PR does the same (so on my mac, I ran this and it's still downloading to /var, not /tmp)
> 
> if that's not convincing, I'll revert back. just lmk!

it's more, the size of the change set scares me. it touches nearly every line of both our probes, which means I'd need to go through every line to ensure I'm comfortable.. even if some are just tabs for a try/catch, it introduces "difference".. this might be good to hang onto in a separate PR that we can toy at though.

makk94:
ive swapped back nocturnal to use a `ts` variable

mireaulf:
Done somewhere else

PR #370, opened 2023-02-21
Adding 107 code

gerbsec:
107 code for exploit attempts


mireaulf:
@gerbsec Is the doc also updated?

gerbsec:
it didnt let me change it which is why i requested the change

PR #376, opened 2023-02-23
Autocomplete for profile

packplusplus:
Auto complete hint for --profile in the cli. If you add `eval "$(_PRELUDE_COMPLETE=bash_source prelude)"` to your bash profile, you will be able to tab 
complete your configured profiles.


privateducky:
closing. reasons:
1. prelude folks are by and large the only users of multiple profiles (for awhile)
2. to use this, you need to know how to add another entry to your bash profile. we already have enough of that.

PR #381, opened 2023-02-24
Update action for verified Friday

bfuzzy1:


privateducky (approved):
this is fine for this week, but next week, we should ensure all demo probes are capable of being uploaded @packplusplus

PR #390, opened 2023-02-27
Prelude interactive - schedule - show test names

mireaulf:


privateducky (approved):
are you just trying to show the test name in this log?
```
print(f'Adding schedule for {tests[test_id]}')
```
if so, we can do so with just: 
```
print(f'Adding schedule for {tests[test_id]}')
```

PR #409, opened 2023-02-28
update codes - python

makk94:


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -50,37 +50,41 @@ def _missing_(cls, value):
...
     @property
     def state(self):
-        for k in ExitCodeGroup:
-            for v in k.value: 
-                if self.value == v.value:
-                    return k
-        return ExitCodeGroup.NONE
-
-
-class ExitCodeGroup(Enum):
```
privateducky, line 60: won't the CLI require a change for this?

makk94: CLI doesn't use ExitCodeGroup that I could find. That's why I said we could just delete it

privateducky: kk perfect; I get the desire, but I want these to be our source of truth.

privateducky (approved):

PR #410, opened 2023-03-01
add plaintext option

makk94:
so everyone can have their precious jq
```
(venv) mahinakaholokula@Mahinas-MacBook-Pro my-prelude % prelude --plaintext detect activity -d 1 | jq
[
  {
    "id": "3645c6a7-7df8-4682-b0e6-ecc3bc9c0ecc",
    "date": "2023-02-28 19:24:50.008086",
    "test": "39de298a-911d-4a3b-aed4-1e8281010a9a",
    "endpoint_id": "beech",
    "status": 100,
    "dos": "linux-x86_64",
    "tags": [
      "tag1"
    ]
  }
]

```

### DIFF
--- a/python/cli/prelude_cli/cli.py
+++ b/python/cli/prelude_cli/cli.py
@@ -12,9 +12,10 @@ from prelude_cli.views.interactive import interactive as interactive_command
 @click.version_option()
 @click.pass_context
 @click.option('--profile', default='default', help='The prelude keychain profile to use', show_default=True)
+@click.option('--plaintext/--no-plaintext', default=False, help='Strip out non-essential output')
 @click.option('--interactive', help='Open interactive wizard (cannot be used with a subcommand)', default=False, is_flag=True)
-def cli(ctx, profile, interactive):
-    ctx.obj = Account(profile=profile)
+def cli(ctx, profile, plaintext, interactive):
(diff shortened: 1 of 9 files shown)


privateducky:
I'm not totally against this... but I am interested in alternative options.
One prime one may be to remove the rich pretty/print of json and revert to Python's normal json printing, which we can do "pretty". the CLI could use a shared function for this pretty print capability, such as:
```
pretty = lambda s: json.dumps(s, indent=4)
```
so we could do something like:
```
print(pretty(json_str))
```

privateducky (changes requested):
new plan...

makk94:
Confused on how this fixes the problem…

privateducky:
> Confused on how this fixes the problem…

I think the pretty print with built in json would work with JQ - but would need to test. If yes, it means we wouldn't need to use an extra flag.

privateducky:
If that doesn't work, I'd say we just avoid pretty printing from the CLI and document that people should use a wrapper like JQ to do it.

makk94:
The issue isn’t the pretty print, it’s your fun little spinner ha

privateducky:
> The issue isn’t the pretty print, it’s your fun little spinner ha

Gotcha. In that case, I won't worry about this for now. You can pipe the output into a file atm.

PR #411, opened 2023-03-01
update CLI w/ decide-recommendation

makk94:


privateducky (changes requested):
funny you're doing this; I spent the a.m. thinking about the API side of this.

I'm looking at what we need to do for "phase 1", ie a social/shared account. to do that, one approach would include splitting the API endpoint into two, one for create, one for update/decision.

but I think its easier to do the social stuff by creating a new "social" account permission (current front runner idea). 

long story short: if we don't separate the recommendation route on the API, let's not separate it here.

long story long: I think you'd prefer this option.. but thoughts on us separating this route by POST (create recommendation) and PUT (make decision)?

makk94:
> long story long: I think you'd prefer this option.. but thoughts on us separating this route by POST (create recommendation) and PUT (make decision)?

lol you do know me. I am not a big fan of our upserts, so I'm on board with separating create and update routes

privateducky (changes requested):
requires rebasing w/rec split

PR #417, opened 2023-03-04
update register endpoint CLI/SDK

makk94:


python/sdk/prelude_sdk/controllers/detect_controller.py
```diff
@@ -10,10 +10,13 @@ def __init__(self, account):
         self.account = account
 
     @verify_credentials
-    def register_endpoint(self, name, tags):
+    def register_endpoint(self, host, serial_num, edr_id, tags):
         """ Register (or re-register) an endpoint to your account """
         with Spinner():
-            params = dict(id=name, tags=tags)
+            id = f'{host}:{serial_num}'
```
privateducky, line 16: since we can default to '' for edr_id, similar to what we do with tags, the API should handle '' as None.

so we can just do:
```
params = dict(id= f'{host}:{serial_num}:{edr_id}', tags=tags)
```

privateducky (changes requested):
1 adjustment

python/sdk/prelude_sdk/controllers/detect_controller.py
makk94: The API does not handle '' as None. You were not fond of the `edr_id or None` clause that would've accounted for that. Should I bring that back? (I can bring it back on either the write to db or read from db, but currently we do write empty string to edr_id and we just read that out and send it directly to user on ListEndpoints)

privateducky (approved):

PR #421, opened 2023-03-06
Update Recommendation type to include events

ceci21:


privateducky (approved):

kenrick:
@ceci21 you will need to bump the pkg version as well

PR #424, opened 2023-03-07
Adding demo4

bfuzzy1:
Pending test ID which I should have from the team today.

EDIT: Test id's added. Pinged @preludeorg/wizards and let Harry know it was ready for review.

### DIFF
--- a/shell/demo/demo4.ps1
+++ b/shell/demo/demo4.ps1
@@ -1,0 +1,172 @@
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory=$true, HelpMessage="Prelude Token")]
+    [String]$preludeToken
+)
+
+function FromEnv { param ([string]$envVar, [string]$default)
+    $envVal = [Environment]::GetEnvironmentVariable($envVar, "User")
+    if ($envVal) { return $envVal } else { return $default }
+}
+
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+$PRELUDE_API = FromEnv "PRELUDE_API" "https://api.preludesecurity.com"
+
+$TEST_ID=""
+$TEST_INTRO="Will your computer quarantine a malicious mach-o file?
+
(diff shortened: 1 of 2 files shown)


shell/demo/demo4.ps1
```diff
@@ -0,0 +1,172 @@
...
+}
+
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+$PRELUDE_API = FromEnv "PRELUDE_API" "https://api.preludesecurity.com"
+
+$TEST_ID="74077d3b-6a2f-49fa-903e-99cad6f34cf6"
+$TEST_INTRO="Will your computer quarantine a malicious mach-o file?
+
+Malicious files are used to gain entry and conduct cyberattacks against corporate systems through seemingly
+innocuous email attachments or direct downloads. For example - a malicious Mac mach-o file was used by Earth Berberoka
```
harry-prelude, line 19: Should we be using `macOS` rather than `Mac`? As far as I can tell, we most commonly use Mac. Just calling to your attention. If there is a strong opinion, lets let it be known.

shell/demo/demo4.ps1
```diff
@@ -0,0 +1,172 @@
...
+during an attack in April 2022.
+
+This test will attempt to download a defanged malicious file to your disk - containing a signature for oRAT used by Earth Berberoka - to see how your endpoint defenses will respond.
```
harry-prelude, line 22: containing a signature for the oRAT malware used by Earth Berberoka

harry-prelude (changes requested):
One clarification, one recommendation.

shell/demo/demo4.ps1
bfuzzy1: fixed.

shell/demo/demo4.ps1
bfuzzy1: I don't think it matters. Consistency is key though so I'll keep Mac.

harry-prelude (approved):
`Mac` it is! Looks good!

privateducky (approved):

PR #425, opened 2023-03-07
add back download CLI cmd

makk94:


mireaulf (approved):
That will also be useful for the sec team

privateducky (approved):

PR #427, opened 2023-03-07
WIP Shell probes no longer use temp only files (PRELUDE_DIR)

privateducky:


privateducky:
@makk94 can you help align raindrop to this change set - so it matches exactly?

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,29 @@
...
-            if test -f "$temp";then
-                $temp
-                dat="${test}:$?"
-                $temp -cleanup
-            else
-                dat="${test}:9"
+        if [ -z "$PRELUDE_CA" ] || [ "$PRELUDE_CA" == "$ca" ];then
+            if test -f $exe;then
+                chmod +x $exe
+                $exe
```
mireaulf, line 21: we don't run cleanup?

privateducky: I'm toying with this now; leaning toward using 'defer' inside our tests instead, to do cleanup after test executions. why? the 'my test was quarantined during execution' edge case makes it a toss up in my mind but defer seems more likely to work in most cases.

mireaulf: Just wanted to make sure you didn't forget about cleanup :)

makk94:
> @makk94 can you help align raindrop to this change set - so it matches exactly?

don't we need the suffix on windows tests to be `.exe` ?

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,29 @@
...
-    test=$(echo $location | grep -o '[0-9a-f]\{8\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{12\}' | head -n 1)
-
-    if [ -z "$test" ];then
-        sleep 14400
-    else
+do  
+    exe=$PRELUDE_DIR/$(uuidgen).vst
+    location=$(curl -sL -w %{url_effective} -o $exe -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "dat:${dat}" $PRELUDE_API)
+    
+    if [ -s $exe ];then
```
makk94, line 14: existence of the file here doesn't mean it's a test in there

ex. this throws any error message in the file too

```
mahinakaholokula@Mahinas-MacBook-Pro my-prelude % curl -sL -w %{url_effective} -o tryme.vst -H "token:itsabadone" -H "dos:darwin-arm64" -H "dat:" "https://api.preludesecurity.com"
https://api.preludesecurity.com/%  
mahinakaholokula@Mahinas-MacBook-Pro my-prelude % cat tryme.vst 
{"Message":"User is not authorized to access this resource"}%            
```

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,29 @@
...
-        if [ -z "$PRELUDE_CA" ] || [ "$PRELUDE_CA" == "$ca" ];then
-            chmod +x $temp
 
-            if test -f "$temp";then
-                $temp
-                dat="${test}:$?"
-                $temp -cleanup
-            else
-                dat="${test}:9"
+        if [ -z "$PRELUDE_CA" ] || [ "$PRELUDE_CA" == "$ca" ];then
```
makk94, line 18: the probe will also spin forever (skipping sleep) if this CA check fails

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,29 @@
...
+            if test -f $exe;then
```
makk94, line 19: should the dat assignment be in this `if` too?

makk94: is this `if` even necessary? You already check the file exists earlier with the `[ -s $exe ] `. Seems like this check here can be combined with that one

shell/probe/nocturnal.sh
privateducky: good point. sticking with $test is probably best. reverted.

shell/probe/nocturnal.sh
privateducky: ah, because it'll re-request. 
that exists today but I'll see if I can revise that in my change set here.

makk94: might I recommend something like this in the while
```
if (test) 
    if (ca passes)
         yay run test
         continue
     fi
fi
clean up files
sleep
```

shell/probe/nocturnal.sh
makk94: i see, guess it's possible it was removed in the very small amount of time between downloading and running.

And the fallback value is always 0 because that comes from the `test -f ` call, right?

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,32 @@
...
+            
+            if test -f $exe;then
+                echo "[P] Running $test masquerading as $exe"
+                $exe
             fi
+            dat="${test}:$?"
         fi
+    else
+        echo "[P] Cleaning up temporary cache"
+        find $PRELUDE_DIR -type f -name "*" -mmin -1 -delete
```
makk94, line 29: don't really need the `-name *` anymore

this doesn't seem particularly scalable (our tests timeout at 10s so if someones setup sucks, they might not finish all tests in 1min. Especially as we keep adding more tests)

also a little scary to "delete all", esp. for a directory as vaguely named as "cache"

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,32 @@
...
-            if test -f "$temp";then
-                $temp
-                dat="${test}:$?"
-                $temp -cleanup
-            else
-                dat="${test}:9"
+        if [ -z "$PRELUDE_CA" ] || [ "$PRELUDE_CA" == "$ca" ];then
+            chmod +x $exe
+            
+            if test -f $exe;then
```
makk94, line 21: nit: use brackets like everywhere else

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,32 @@
...
+dos=$(echo $(uname -s)-$(uname -m))
 
 while :
-do
-    temp=$(mktemp)
-    location=$(curl -sL -w %{url_effective} -o $temp -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "dat:${dat}" $PRELUDE_API)
-    test=$(echo $location | grep -o '[0-9a-f]\{8\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{12\}' | head -n 1)
+do  
+    exe=$PRELUDE_DIR/$(uuidgen)
+    location=$(curl -sL -w %{url_effective} --create-dirs -o $exe -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "dat:${dat}" $PRELUDE_API)
```
makk94, line 12: am i crazy or do we need to reset dat to empty string when we sleep?

I don't see this in our code today, so hoping I'm crazy...

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,32 @@
 #!/bin/bash
 
 PRELUDE_API=${PRELUDE_API:="https://api.preludesecurity.com"}
+PRELUDE_DIR=${PRELUDE_DIR:="./cache"}
+PRELUDE_SLEEP=${PRELUDE_SLEEP:=14440}
 
-sys=$(uname -s)-$(uname -m)
-dos=$(echo $sys | tr '[:upper:]' '[:lower:]')
+dos=$(echo $(uname -s)-$(uname -m))
```
makk94, line 7: dont need the `echo` bit

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,33 @@
 #!/bin/bash
 
+PRELUDE_SLEEP=${PRELUDE_SLEEP:=14440}
+PRELUDE_DIR=${PRELUDE_DIR:="./cache"}
 PRELUDE_API=${PRELUDE_API:="https://api.preludesecurity.com"}
+PRELUDE_CA=${PRELUDE_CA:="prelude-account-prod-us-west-2.s3.amazonaws.com"}
```
makk94, line 6: arent we us-west-1

shell/probe/nocturnal.sh
privateducky: yeah; I've been working on this area a bit. going between a few thoughts atm.

shell/probe/nocturnal.sh
privateducky: good question.. I'll look into that (both pre/post this change)

shell/probe/nocturnal.sh
makk94: why are you keeping the echo ?

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,32 @@
 #!/bin/bash
 
-PRELUDE_API=${PRELUDE_API:="https://api.preludesecurity.com"}
+PRELUDE_SLEEP=${PRELUDE_SLEEP:=14440}
+PRELUDE_DIR=${PRELUDE_DIR:="./cache"}
+PRELUDE_CA=${PRELUDE_CA:="prelude-account-prod-us-west-1.s3.amazonaws.com"}
 
-sys=$(uname -s)-$(uname -m)
-dos=$(echo $sys | tr '[:upper:]' '[:lower:]')
+api="https://api.preludesecurity.com"
```
makk94, line 7: pleeease keep the ability to override for staging

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,32 @@
...
 while :
-do
-    temp=$(mktemp)
-    location=$(curl -sL -w %{url_effective} -o $temp -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "dat:${dat}" $PRELUDE_API)
-    test=$(echo $location | grep -o '[0-9a-f]\{8\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{12\}' | head -n 1)
+do  
+    exe=$PRELUDE_DIR/$(uuidgen)
+    location=$(curl -sL -w %{url_effective} --create-dirs -o $exe -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "dat:${dat}" $api)
+    test=$(echo $location | grep -o '[0-9a-f]\{8\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{4\}-[0-9a-f]\{12\}' | head -n 1)    
+    unset dat
```
makk94, line 15: this unset should be near the sleep, no?

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,32 @@
...
-            else
-                dat="${test}:9"
-            fi
+        if [ "$PRELUDE_CA" == "$ca" ];then
+            echo "[P] Running $test masquerading as $exe"
+            chmod +x $exe
+            $exe
+            dat="${test}:$?"
+        else
+            echo "[P] Authority mismatch: $exe" && exit 0
```
makk94, line 26: wouldn't you rather print $ca ?

shell/probe/nocturnal.sh
privateducky: are you looking at an old commit? there are only 3 echo statements currently in this PR, all debug statements that may remain.

shell/probe/nocturnal.sh
privateducky: good point; will update

shell/probe/nocturnal.sh
privateducky: could go either way; not sure I care but will examine both

shell/probe/nocturnal.sh
privateducky: downside is someone can use an env var to override where tests are coming from, so I'll likely opt not to keep that in.

shell/probe/nocturnal.sh
privateducky: good call

makk94: have you maybe not committed what's on your local?

I'm seeing this
Screenshot 2023-03-07 at 4 53 31 PM

shell/probe/nocturnal.sh
makk94: isn't that what the CA check is for?

shell/probe/nocturnal.sh
makk94: thing I'm thinking about that needs the unset near the sleep:

iteration 1:
 - get a test, set dat={123-123-123:100}
 iteration 2:
 - send in that dat and get back "no more tests"
 - go to sleep
 iteration 3:
- exit sleep, dat still set to {123-123-123:100}
- send in that dat while getting next round of tests

shell/probe/nocturnal.sh
privateducky: sort of... two different things.. but the CA check may override the concern as you're pointing out... I think you're spot on

shell/probe/nocturnal.sh
makk94: you're looking at an old comment ;)

shell/probe/nocturnal.sh
privateducky: on iteration 2, when you get 'no more tests', you'd be unsetting the 'data' before getting to the if/check on the $test.

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,33 @@
 #!/bin/bash
 
+PRELUDE_SLEEP=${PRELUDE_SLEEP:=14440}
+PRELUDE_DIR=${PRELUDE_DIR:="./cache"}
```
makk94, line 4: could we at least call this somethign more prelude-specific?

shell/probe/nocturnal.sh
privateducky: (but I swapped to sleep anyway, as I think it reads better)

shell/probe/nocturnal.sh
privateducky: yea, still thinking on this new variable overall.. got a name suggestion vs cache?

shell/probe/nocturnal.sh
```diff
@@ -1,30 +1,33 @@
...
-dos=$(echo $sys | tr '[:upper:]' '[:lower:]')
+dos=$(uname -s)-$(uname -m)
 
 while :
-do
-    temp=$(mktemp)
-    location=$(curl -sL -w %{url_effective} -o $temp -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "dat:${dat}" $PRELUDE_API)
+do  
+    exe=$PRELUDE_DIR/$(uuidgen)
+    location=$(curl -sL -w %{url_effective} --create-dirs -o $exe -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "dat:${dat}" $PRELUDE_API)
```
makk94, line 13: will you make the curl flags include `curl -sfSL` so it'll print out any http error codes

shell/probe/nocturnal.sh
makk94: im bad at names ha but presumably something including the work "prelude" unless there's a reason you're avoiding it

shell/probe/nocturnal.sh
privateducky: hmm, where would those print to in that case, as we're mapping to $location?

makk94: the -f has the  http failure code print to stdout. location is still filled with url_effective

the -S isn't as necessary now that we're deleting the files, but that stop the file from being created on http error code. The --create-dirs still takes effect tho

shell/probe/nocturnal.sh
makk94: why'd this go back :(

privateducky: can't leave API and CA both mutable

PR #429, opened 2023-03-08
Ensuring raindrop is aligned w/nocturnal changes

privateducky:


shell/probe/raindrop.ps1
```diff
@@ -1,46 +1,43 @@
...
     Param([String]$Dat = "")
 
-    $TempFile = [System.IO.Path]::GetTempFileName() | Rename-Item -NewName { [System.IO.Path]::ChangeExtension($_, "exe") } -PassThru
-
-    $Headers = @{
-        'token' = $Token
-        'dos' = $Dos
-        'dat' = $Dat
-    }
     try {
```
mireaulf, line 6: You can get rid of the try/catch

mireaulf (commented):
Where do we delete the vst?

shell/probe/raindrop.ps1
privateducky: are you confident nothing could throw an unhandled exception in that block?

privateducky:
> Where do we delete the vst?

good eye; just added

shell/probe/raindrop.ps1
```diff
@@ -1,46 +1,39 @@
...
-        $Response = Invoke-WebRequest -URI $Address -UseBasicParsing -Headers $Headers -MaximumRedirection 1 -OutFile $TempFile -PassThru
-    } catch {
-        $StatusCode = $_.Exception.Response.StatusCode.value__
-        Write-Output "ERROR: Failed to reach Prelude Service. " + $StatusCode
-        return
-    }
-    if ($CA -and $CA -ne $Response.BaseResponse.ResponseUri.Authority) {
-        return
+    $Response = Invoke-WebRequest -URI $Api -UseBasicParsing -Headers $Headers -MaximumRedirection 1 -OutFile $Vst -PassThru
+    if ($CA -ne $Response.BaseResponse.ResponseUri.Authority) {
```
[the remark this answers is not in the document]
privateducky: so we should flip and look at the test first, right?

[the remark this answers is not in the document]
privateducky: updated

shell/probe/raindrop.ps1
```diff
@@ -1,46 +1,39 @@
...
+    Run -Dat "${Test}:$Code"
 }
 
 function Execute { 
     Param([String]$File)
+
     try {
-        return (Start-Process -FilePath $File -Wait -NoNewWindow -PassThru).ExitCode
+        $Code = (Start-Process -FilePath $File -Wait -NoNewWindow -PassThru).ExitCode
+        return If (Test-Path $Vst) {$Code} Else {127}
```
makk94, line 31: question: does it count as PROTECTED if the file was removed after it was already run?

makk94: also should "$Vst" here be "$File" ?

shell/probe/raindrop.ps1
```diff
@@ -1,46 +1,39 @@
...
+
 function Run {
     Param([String]$Dat = "")
 
-    $TempFile = [System.IO.Path]::GetTempFileName() | Rename-Item -NewName { [System.IO.Path]::ChangeExtension($_, "exe") } -PassThru
-
+    $Vst = New-Item -Path "$Dir\$(New-Guid).exe" -Force
     $Headers = @{
-        'token' = $Token
+        'token' = FromEnv "PRELUDE_TOKEN"
```
mireaulf, line 8: Since we read the token every time, if the token changes in the env different runs will beacon with different tokens

shell/probe/raindrop.ps1
```diff
@@ -50,13 +43,17 @@ function FromEnv { param ([string]$envVar, [string]$default)
...
+$Api = "https://api.preludesecurity.com"
+$Dos = "windows-$Env:PROCESSOR_ARCHITECTURE" 
 
 while ($true) {
-    Run
-    Start-Sleep -Seconds 14400
-}
+    try {
+        Run
+        Get-ChildItem -Path $Dir -Include * | Where-Object{$_.LastWriteTime -gt (Get-Date).AddMinutes(-5)}| Remove-Item
```
mireaulf, line 56: That's more or less the same behavior as nocturnal but I buy it

shell/probe/raindrop.ps1
makk94, line 31: return If throws an error. Either separate the lines or it probably works if you make it a sub-expression

makk94: @privateducky first question still stands. Still think that if the file runs, and then its deleted after the run, it's UNPROTECTED, not QUARANTINED

shell/probe/raindrop.ps1
privateducky: yep; we do the same thing in nocturnal (by design) -- so we can modify the value on the fly, if needed

shell/probe/raindrop.ps1
privateducky: we mark it as 127 (quarantined) if the file no longer exists after running it. if something on the system removed the file, I wouldn't mark that as unprotected, I'd mark that as something quarantined our test.

privateducky: How sure are you?

Screenshot 2023-03-08 at 3 46 39 PM

makk94: if it ran successfully doesn't that mean the bad thing already happened?  If it was quarantined after that then future runs of the test can say PROTECTED, but I'd think that the run that finished would be considered UNPROTECTED

makk94: you've just taken the return out, which is the problem part, and showed me the sub-expression works..

Have you run this code path?

PR #433, opened 2023-03-08
upload as binary files

makk94:
Re: https://detectors.atlassian.net/browse/DET-45

relies on https://github.com/preludeorg/service/pull/686

Testing (personal stack). Created a new test and uploaded some files to it. Also downloaded all  of them back to local to confirm text is correct/ executables run. Here's my s3 of what was tested:

Screenshot 2023-03-09 at 8 36 06 AM


mireaulf:
Does that behave correctly for non binary data?

makk94:
> Does that behave correctly for non binary data?

Just noticed that we upload on creates - so there's an example of sending through plain text (note we don't set content-type to application/octet-stream). And then we do set that header on the upload command b/c that's always sent through as bytes

mireaulf (approved):

privateducky (approved):

PR #437, opened 2023-03-09
Add machine flag to set env var

mireaulf:


shell/probe/README.md
```diff
@@ -9,7 +9,7 @@ Shell probes are intended to be dependency-free executable code that works on mo
 ### Raindrop 
 
 ```terminal
-SETX PRELUDE_TOKEN <TOKEN>
+SETX PRELUDE_TOKEN <TOKEN> /M
```
privateducky, line 12: is this how the installers are doing this as well?

mireaulf: Now they do

privateducky: LOL that's good.
I was actually referring to the new installers though haha

mireaulf: The installers don't really touch the token right?

mireaulf: Installers are using machine!

privateducky: Right. Are they setting it the same way as you're proposing here?

mireaulf: They are
https://github.com/preludeorg/systems/blob/e2bd451adfd149426d190e51ca35b361b23ccaae/scripts/installers/windows/Detect/Services/DetectService.cs#L37

privateducky: perfect.

I was actually looking at this area yesterday. before we commit to this.. is there a way we can check all env var "locations" simply enough to avoid any confusion?

I ask, because I was originally trying raindrop on my Mac, with PowerShell core. There are only 2 incompatible lines of code, this is one of them. the line works if I don't specify 'machine' or 'user' though.

mireaulf: The process is supposed to inherit the value from the machine and user so more testing would be needed to figure it out but we could probably get rid of that (in a future PR)

privateducky: agreed

privateducky (approved):

PR #445, opened 2023-03-14
print curl fails

makk94:


shell/probe/nocturnal.sh
```diff
@@ -10,7 +10,7 @@ dos=$(uname -s)-$(uname -m)
 while :
 do  
     exe=$PRELUDE_DIR/$(uuidgen)
-    location=$(curl -sL -w %{url_effective} --create-dirs -o $exe -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "dat:${dat}" $api)
+    location=$(curl -sfSL -w %{url_effective} --create-dirs -o $exe -H "token:${PRELUDE_TOKEN}" -H "dos:${dos}" -H "dat:${dat}" $api)
```
privateducky, line 13: can you post what this will show?

makk94: w/ a bad token:
```
mahinakaholokula@Mahinas-MacBook-Pro libraries % ./shell/probe/nocturnal.sh 
curl: (22) The requested URL returned error: 403
```

w/o the `-fS` flags, there's no output at all

mireaulf: -S: show errors
-f: return error 22 instead of a html document

mireaulf:
Why that change in the first place?

PR #449, opened 2023-03-15
Send up current probe version w/each request

privateducky:


makk94:
1. do you want to update the demo scripts or do they not matter?
2. we're not using this param in the BE yet - is this PR just getting ready for some future plans?

makk94 (approved):

mireaulf:
If we wanted to go this way, I'd probably look into sending version in the headers

PR #450, opened 2023-03-15
Update an account mode from cli

privateducky:


python/sdk/prelude_sdk/controllers/iam_controller.py
```diff
@@ -32,6 +32,15 @@ def purge_account(self):
                 raise Exception(res.text)
             return res.text
 
+    @verify_credentials
+    def update_account(self, mode: int):
+        """ Update properties on an account """
+        with Spinner():
+            params = dict(mode=mode)
+            res = requests.put(f'{self.account.hq}/iam/control', headers=self.account.headers, json=params)
```
makk94, line 40: /iam/account

python/cli/prelude_cli/views/iam.py
```diff
@@ -25,6 +25,18 @@ def register_account(controller):
     print_json(data=creds)
 
 
+@iam.command('update-account')
+@click.option('-m', '--mode',
+              help='provide a mode',
+              default='manual', show_default=True,
+              type=click.Choice(['manual', 'frozen'], case_sensitive=False))
```
makk94, line 32: if we might add more modes, I'd dynamically create this list from the enum like we do elsewhere

python/cli/prelude_cli/views/iam.py
```diff
@@ -25,6 +25,18 @@ def register_account(controller):
...
+@click.pass_obj
+@handle_api_error
+def update_account(controller, mode):
+    """ Update an account """
+    controller.update_account(mode=mode)
```
makk94, line 37: sdk param is listed as an int, what's this?

python/cli/prelude_cli/views/iam.py
privateducky: Agreed

python/sdk/prelude_sdk/controllers/iam_controller.py
privateducky: Good eye

python/cli/prelude_cli/views/iam.py
privateducky: That's me flipping between a string, an enum and an int, followed by you having a better eye than me!

PR #461, opened 2023-03-17
Use event log if possible

mireaulf:


shell/probe/raindrop.ps1
```diff
@@ -1,5 +1,11 @@
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 
+function Log {
+    Param([String]$line)
+
+    Write-EventLog -LogName "Application" -Source "Prelude Probe Service" -EventId 0 -Message $line -ErrorAction Ignore
```
privateducky, line 6: Why a separate function vs just using this instead of write output?

mireaulf: Using this Log function = 2003 chars
Copy pasting that same line = 2156 chars

privateducky: Good

privateducky (approved):

PR #462, opened 2023-03-20
lowercase endpoint_id filters so filter is not case sensitive

makk94:
fixes: https://detectors.atlassian.net/browse/DET-21


makk94:
https://github.com/preludeorg/service/pull/711

PR #478, opened 2023-03-24
New Demo Action

packplusplus:


privateducky (changes requested):
Let's also delete all existing demo scripts

privateducky (approved):

PR #484, opened 2023-03-27
move around the $dat

privateducky:


khyberspache:
Dat never gets `unset` in this scheme - you will continually send `test:<code>` to the API with every request

PR #485, opened 2023-03-27
Update the demo probe to notify if the health check was quarantined o…

khyberspache:
…r blocked

image


privateducky (changes requested):
note that there are 2 health checks in the tests now, not just the 1

makk94:
Could you not also print out the "control test passed" - think that's confusing to see since I'm not sure if "passing" means quarantined or not anymore

Also if you want to get fancy, you could update the summary at the end of the script to not print "PROTECTED" for the test

privateducky (approved):

PR #496, opened 2023-03-31
feat: interactive open dashboard with credentials

mireaulf:
Depends on https://github.com/preludeorg/platform/pull/870

Closes https://detectors.atlassian.net/browse/DET-100

### DIFF
--- a/python/cli/prelude_cli/views/interactive.py
+++ b/python/cli/prelude_cli/views/interactive.py
@@ -1,6 +1,8 @@
 import os
+import json
 import uuid
 import click
+import base64
 import shutil
 import socket
 import webbrowser
@@ -874,9 +876,9 @@ class ExecutiveDashboard:
         self.title = title
 
     def enter(self):
-        account = self.wiz.account.headers['account']
-        token = self.wiz.account.headers['token']
-        webbrowser.open(f'https://platform.preludesecurity.com?account={account}&token={token}', new=2)
+        d = dict(account=self.wiz.account.headers['account'], token=self.wiz.account.headers['token'])
+        data = base64.urlsafe_b64encode(json.dumps(d).encode('utf8')).decode('utf8')
+        webbrowser.open(f'https://platform.preludesecurity.com#{data}', new=2)
 
 
 @click.command()
--- a/python/cli/prelude_cli/views/interactive.py
+++ b/python/cli/prelude_cli/views/interactive.py
@@ -1,8 +1,6 @@
 import os
-import json
 import uuid
 import click
-import base64
 import shutil
 import socket
 import webbrowser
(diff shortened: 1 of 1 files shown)


python/cli/prelude_cli/views/interactive.py
```diff
@@ -873,7 +874,9 @@ def __init__(self, wiz: Wizard, title: str):
         self.title = title
 
     def enter(self):
-        webbrowser.open('https://platform.preludesecurity.com/detect', new=2)
+        account = self.wiz.account.headers['account']
+        token = self.wiz.account.headers['token']
+        webbrowser.open(f'https://platform.preludesecurity.com?account={account}&token={token}', new=2)
```
privateducky, line 879: @kenrick thoughts?

kenrick: Yeah I asked @mireaulf to use `#` instead of a query string `?` this way the browser will not keep the account and token in its history or send it the server

python/cli/prelude_cli/views/interactive.py
```diff
@@ -873,7 +876,9 @@ def __init__(self, wiz: Wizard, title: str):
         self.title = title
 
     def enter(self):
-        webbrowser.open('https://platform.preludesecurity.com/detect', new=2)
+        d = dict(account=self.wiz.account.headers['account'], token=self.wiz.account.headers['token'])
+        data = base64.urlsafe_b64encode(json.dumps(d).encode('utf8')).decode('utf8')
+        webbrowser.open(f'https://platform.preludesecurity.com#{data}', new=2)
```
privateducky, line 881: we should use the ``hq`` entry from the keychain

python/cli/prelude_cli/views/interactive.py
```diff
@@ -54,6 +56,7 @@ class Wizard:
 
     def __init__(self, account):
         # SDK controllers
+        self.account = account
```
privateducky, line 59: this property is an outlier here.
it is under the title/comment "SDK controllers" but it is not that.

python/cli/prelude_cli/views/interactive.py
mireaulf: Unless we strip the `api.` part of the url

privateducky: Question: what if we instead hit the verify user route on Service (assuming we adjust it to always do a redirect, even on recertification)? Doing that could open the browser with the right account.. would that work?

mireaulf:
Depends on https://github.com/preludeorg/service/pull/766

privateducky (approved):

python/cli/prelude_cli/views/interactive.py
```diff
@@ -873,7 +873,9 @@ def __init__(self, wiz: Wizard, title: str):
         self.title = title
 
     def enter(self):
-        webbrowser.open('https://platform.preludesecurity.com/detect', new=2)
+        account = self.wiz.detect.account.headers['account']
+        token = self.wiz.detect.account.headers['token']
+        webbrowser.open(f'https://{self.wiz.detect.account.hq}/iam/user?account={account}&token={token}', new=2)
```
makk94, line 878: will account.hq not already have the https:// in it?

mireaulf: Correct; opening a PR
https://github.com/preludeorg/libraries/pull/519

PR #497, opened 2023-03-31
feat: do not pass test_id

mireaulf:
Depends on https://github.com/preludeorg/service/pull/756

Closes https://detectors.atlassian.net/browse/DET-22

### DIFF
--- a/javascript/sdk/index.html
+++ b/javascript/sdk/index.html
@@ -118,11 +118,12 @@
           try {
             service = await setupAccount();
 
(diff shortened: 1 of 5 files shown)


makk94:
Don't we use this field? To allow updating an existing test / so Patrick can upload Prelude tests w/ our pre-determined test IDs

mireaulf:
This is `create_test`, `upload_attachment` still exists to upload after creating a test

PR #504, opened 2023-04-06
move keychain/spinner from SDK to CLI

makk94:


python/cli/prelude_cli/views/interactive.py
```diff
@@ -909,9 +909,16 @@ def interactive(account):
...
                 break
 
-            wizard.iam.new_account(handle=os.getlogin())
-            keychain = PurePath(Path.home(), '.prelude', 'keychain.ini')
-            print(f'Account created! Credentials are stored in your keychain: {keychain}')
+            email = Prompt.ask('Enter your email handle')
+            creds = wizard.iam.new_account(handle=email)
+
+            cfg = ctx['keychain'].read_keychain_config()
+            cfg[ctx['keychain'].profile]['account'] = creds['account_id']
```
mireaulf, line 916: Could we pass around an object instead of a dict for ctx here?

python/cli/prelude_cli/views/iam.py
```diff
@@ -12,18 +13,24 @@
...
-    ctx.obj = IAMController(account=ctx.obj)
+    ctx.obj['controller'] = IAMController(account=ctx.obj['account'])
 
 
 @iam.command('create-account')
 @click.pass_obj
 @handle_api_error
 @click.confirmation_option(prompt='Overwrite local account credentials for selected profile?')
-def register_account(controller):
+def register_account(ctx):
```
mireaulf, line 23: Same question here

python/cli/prelude_cli/views/interactive.py
makk94: we couuld, but I am not creative enough to see a good object to pass through. The things this context object needs at some point or another:
- the SDK account object, the CLI Keychain object (both set in cli.py), the SDK controller object (set in each controller)

It's that last one that makes a single object not work perfectly for me

makk94: I suppose I could make a CLI Account object that inits w/ the SDK account and CLI Keychain and then add a get/set_controller() type of functionality to it...

sounds a little overkill to me, but can try it if you think it sounds better

PR #506, opened 2023-04-07
add option to pass through account/token to SDK constructor

makk94:
if you provide account/token to SDK Account initializer, use that instead of reading/updating a keychain file


privateducky:
too heavy handed; I just merged an option from @khyberspache that approaches keychain interactions nicely

PR #509, opened 2023-04-07
Update JS SDK Version

bfuzzy1:


packplusplus:
We already shipped the needed change as part of 58, and the change that needs to happen is https://github.com/preludeorg/platform/pull/879/files

PR #525, opened 2023-04-11
Add cucumber testing for demo experience

ceci21:


tests/requirements.txt
```diff
@@ -0,0 +1,2 @@
+behave == 1.2.6
+prelude-sdk == 0.9.94
```
mireaulf, line 2: Don't pin the sdk if we want to run this continuously

PR #537, opened 2023-04-13
New user with service permission don't need to verify email

mireaulf:


python/cli/prelude_cli/views/iam.py
```diff
@@ -62,7 +62,8 @@ def create_user(controller, permission, handle, days):
         expires=expires
     )
     print_json(data=resp)
-    print("\nCheck your email to verifiy your account.\n")
+    if permission != Permission.SERVICE.name:
+        print("\nCheck your email to verifiy your account.\n")
```
makk94, line 66: thought i fixed this already, but spelling on verify please

PR #558, opened 2023-04-21
Delete vst on every loop no matter what

khyberspache:


shell/probe/raindrop.ps1
```diff
@@ -52,11 +52,11 @@ while ($true) {
...
         } else {
             $Dat = ""
-            Remove-Item $Dir -Force -Recurse
             Start-Sleep -Seconds $Sleep
         }
     } catch { 
         Write-Error $_
         Start-Sleep -Seconds $Sleep
     }
+    Remove-Item $Dir -Force -Recurse -ErrorAction SilentlyContinue
```
privateducky, line 61: this will delete the directory after each test vs every test loop/iteration. thoughts on this change?

makk94: is it still a requirement that raindrop mimic nocturnal exactly?

mireaulf: yes

mireaulf: > this will delete the directory after each test vs every test loop/iteration. thoughts on this change?

I don't like that it's done after the sleep


> is it still a requirement that raindrop mimic nocturnal exactly?

It should

khyberspache: It's the root cause of this https://detectors.atlassian.net/browse/DET-299?atlOrigin=eyJpIjoiYjlkZjM2ZWI0YTk2NGEwZWIwYjg2M2M5MGJmYjM0NTAiLCJwIjoiamlyYS1zbGFjay1pbnQifQ

khyberspache: I can move it before the sleeps, but the deletion MUST happen no matter what. We are currently not ensuring that and leaving artifacts on the system

khyberspache: Pushed a commit to delete before sleep

makk94 (approved):

shell/probe/raindrop.ps1
khyberspache: > is it still a requirement that raindrop mimic nocturnal exactly?

We would need to bring Nocturnal in line if this change is accepted @makk94

makk94: nocturnal looks good as is

mireaulf (approved):

shell/probe/raindrop.ps1
```diff
@@ -52,11 +52,12 @@ while ($true) {
             exit 1
         } else {
             $Dat = ""
-            Remove-Item $Dir -Force -Recurse
+            Remove-Item $Dir -Force -Recurse -ErrorAction SilentlyContinue
```
privateducky, line 55: what if we put this logic into the catch block and raise an exception here in the else block? I like that because it's duplicate logic that we'll want (it'd be nice to drop the $Dat in the catch as well). 

this also gives us a nice print statement after each test loop.

khyberspache: Love it

privateducky (approved):

PR #565, opened 2023-04-28
New cli

packplusplus:


privateducky (changes requested):
we need to also update the SDK now.

privateducky:
needs to bump to 1.1 and include all pieces: JS SDK, py SDK, py CLI

PR #566, opened 2023-04-29
shifting route of partners API (attach/detach)

privateducky:
Requires test updates.

### DIFF
--- a/python/sdk/prelude_sdk/controllers/partner_controller.py
+++ b/python/sdk/prelude_sdk/controllers/partner_controller.py
@@ -15,7 +15,7 @@ class PartnerController:
         with Spinner():
             params = dict(api=api, user=user, secret=secret)
             res = requests.post(
-                f'{self.account.hq}/iam/partner/{name}',
+                f'{self.account.hq}/partner/attach/{name}',
                 headers=self.account.headers,
                 json=params,
                 timeout=10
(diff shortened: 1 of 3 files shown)


python/sdk/prelude_sdk/controllers/partner_controller.py
```diff
@@ -15,7 +15,7 @@ def attach(self, name: str, api: str, user: str, secret: str):
         with Spinner():
             params = dict(api=api, user=user, secret=secret)
             res = requests.post(
-                f'{self.account.hq}/iam/partner/{name}',
+                f'{self.account.hq}/partner/attach/{name}',
```
makk94, line 18: not sure we need the "attach"/"detach" in the path. the http request type already tells us that

privateducky: we already have a GET/POST for /partner.. but I'm open to alternatives!

makk94: imo POST/DELETE should be for attach/detach, and deploy/get_endpoints should be the more custom paths. 

i'd go with POST /partner/deploy and GET /partner/endpoints (I think GET /partner would return a list of partners, not a list of endpoints on the partner). these switches require re-orging our lambdas, but that's what i'd do

privateducky: I'm good w/that. 
can you post a set of PRs with the change ahead of tomorrow's upgrade?

PR #574, opened 2023-05-01
update sdk versioning for rule to unit change

stephanwampouille:


privateducky (changes requested):
Move this to v 1.1.0, as it is a feature upgrade

stephanwampouille:
> Move this to v 1.1.0, as it is a feature upgrade

done

privateducky (approved):

PR #582, opened 2023-05-03
add optional name/company props on register_account

makk94:
re: https://github.com/preludeorg/service/pull/860


python/cli/prelude_cli/views/iam.py
```diff
@@ -22,8 +22,12 @@ def iam(ctx):
 def register_account(controller):
     """ Register a new account """
     handle = click.prompt('Enter a handle')
```
privateducky, line 24: This should specify email

privateducky (changes requested):

privateducky:
Closed; we'll need to revisit this topic this week - but tie the names to the users vs account.

PR #586, opened 2023-05-03
fix: change code to alert

kenrick:


mireaulf (approved):

javascript/sdk/lib/types.ts
```diff
@@ -18,7 +18,7 @@ export interface Test {
   id: string;
   name: string;
   unit: string;
-  code?: string;
+  alert: string;
```
makk94, line 21: do you still want the question mark?

privateducky (approved):

javascript/sdk/lib/types.ts
kenrick: @mireaulf says its not optional just empty (`""`)  when blank

PR #596, opened 2023-05-07
add bypass method for account creation

khyberspache:


python/cli/prelude_cli/views/iam.py
```diff
@@ -18,12 +18,16 @@ def iam(ctx):
 @iam.command('create-account')
 @click.pass_obj
 @handle_api_error
+@click.option('-m', '--method',
+              help='provide a registration method',
+              default='email', show_default=True,
+              type=click.Choice(['email', 'bypass'], case_sensitive=False))
```
privateducky, line 24: providing options seems unnecessary. couldn't we just make the click.option a flag that defaults bypass=False?

doing that, I'd avoid sending the 'method' in the body but instead use it as a query parameter, so you could do something like ``POST:/iam/account?bypass=1``

privateducky: Unless you're tying this in somehow in the OAUTH version... I think either way though, I'd opt for the method being in the query params vs the body, as it's a property to direct the API toward a specific path... so ``/iam/account?method=bypass`` could be an alternative option.

khyberspache: updated to the latter as a query param

python/cli/prelude_cli/views/iam.py
```diff
@@ -18,12 +18,12 @@ def iam(ctx):
...
-def register_account(controller):
+@click.option('-b', '--bypass', help='bypass email verification', is_flag=True, default=False)
+def register_account(controller, bypass):
     """ Register a new account """
     handle = click.prompt('Enter a handle')
     with Spinner():
-        data = controller.new_account(handle=handle)
+        data = controller.new_account(handle=handle, bypass=bypass)
     print_json(data=data)
     print("\nCheck your email to verify your account.\n")
```
privateducky, line 28: Let's adjust the msg here, to not direct the user to check their email if bypassing. While you're in there, change the click option from handle to email.

privateducky (changes requested):

privateducky (approved):

PR #598, opened 2023-05-07
adding a pylint workflow

khyberspache:


.github/workflows/pylint.yml
```diff
@@ -0,0 +1,34 @@
+name: Pylint
```
privateducky, line 1: if we added another language, like JS, to the mix here.. what would that look like?

.github/workflows/pylint.yml
```diff
@@ -0,0 +1,34 @@
...
+    - uses: actions/checkout@v3
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install pylint
+    - name: Analysing the code with pylint
```
privateducky, line 21: if we made the 'dotfiles' repo public, could we clone here instead, to create consistency?

khyberspache: yep would work fine

.github/workflows/pylint.yml
khyberspache: we would add either:
a) another workflow yml
b) break this into 2 jobs and run one against all the JS files

privateducky:
closing to resume this work at some point in the future.

PR #600, opened 2023-05-08
update demo script (test only LockBit, update PROTECTED codes)

makk94:
tested on my mac and a windows 2022 VM

### DIFF
--- a/shell/demo/demo.ps1
+++ b/shell/demo/demo.ps1
@@ -4,6 +4,7 @@ param(
     [String]$preludeToken
 )
 
+$ProgressPreference = 'SilentlyContinue'
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 $PRELUDE_API = if ($env:PRELUDE_API) { $env:PRELUDE_API } else { "https://api.preludesecurity.com" };
 $PRELUDE_DIR = if ($env:PRELUDE_DIR) { $env:PRELUDE_DIR } else { ".vst" };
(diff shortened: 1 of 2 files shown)


shell/demo/demo.ps1
```diff
@@ -8,17 +8,19 @@ param(
...
 }
 
 $Tests = [ordered]@{
-    "39de298a-911d-4a3b-aed4-1e8281010a9a" = "Health check"
-    "3ebbda49-738c-4799-a8fb-206630cf609e" = "Will a long running VST be stopped properly?"
-    "2e705bac-a889-4283-9a8e-a12358fa1d09" = "Will your computer quarantine Royal Ransomware?"
-    "b74ad239-2ddd-4b1e-b608-8397a43c7c54" = "Will your computer quarantine a malicious Office document?"
-    "ca9b22be-93d5-4902-95f4-4bc43a817b73" = "Will your computer quarantine Colour-Blind malware?"
+    "39de298a-911d-4a3b-aed4-1e8281010a9a" = "Health Check"
+    "3ebbda49-738c-4799-a8fb-206630cf609e" = "Long Running VST"
```
privateducky, line 19: let's not include 'long running VST'... I think this test may not exist soon...

privateducky (changes requested):
@makk94 I think you misunderstood me actually...
but it's ok, because we had to do this anyway this week 

1. the demo_account function on Service is the one that needs the emerging threat included
2. these demo scripts should be trimmed down to ONLY include the LockBit (NGAV) test.

sorry for the confusion but at least you're knocking both things out at once!

makk94:
Ohhhh
lol. too many demo things. k, i'll go update the demo account

any reason we're moving back to this demo being a single test instead a list of 5?

privateducky:
> Ohhhh lol. too many demo things. k, i'll go update the demo account
> 
> > 2. these demo scripts should be trimmed down to ONLY include the LockBit (NGAV) test.
> 
> (readme says this is AV not NGAV btw) Screenshot 2023-05-08 at 12 35 07 PM
> 
> any reason we're moving back to this demo being a single test instead a list of 5? (is this a long term decision/ should I rewrite demo to assume single test again)

1. Short term move to 1 test, as we're testing out some new things.
2. We have 2 LockBit tests, one for AV and one for NGAV. The demo should be the latter.

privateducky (approved):

privateducky:
@makk94 great/clean work. let me know once you've confirmed against staging

makk94:
I tested this against staging already. unless you mean teh full flow (ie. i need to bother patrick to update demo in s3 and then test from download)?

PR #603, opened 2023-05-09
Migrate Endpoint

bfuzzy1:
Staging this, but we will be migrating our Endpoint package here per David's request.


privateducky (commented):
@bfuzzy1 why is this marked as draft?

I think what I'd like to see here is a directory structure of:
```
go/tests/**
```

This would extend the concept of our libraries repo to be a) probes b) SDKs c) tests. Each could come in a variety of languages, which would play well as we open VSTs to new languages in the future (basically, just adding support to Compute). 

We should rework the README in this PR to clearly define these 3 elements.

privateducky (approved):

PR #608, opened 2023-05-10
add expires to User object

stephanwampouille:
Is this all we have / need so far @makk94


makk94:
yep, that's all we have for ya
comes back as a string formatted like this right now `"expires": "2024-05-09 14:40:32.510834"`

kenrick:
Looks good @stephanwampouille

makk94 (approved):

PR #616, opened 2023-05-12
fix: update SDK tests to create accounts with name

kenrick:


packplusplus (approved):
Looks good. Thank you!

PR #624, opened 2023-05-16
update version and default CA

khyberspache:


shell/probe/raindrop.ps1
```diff
@@ -23,33 +23,37 @@ function FromEnv { param ([string]$envVar, [string]$default)
...
-        $Vst = New-Item -Path "$Dir\$(New-Guid).exe" -Force
         $Headers = @{
             'token' = FromEnv "PRELUDE_TOKEN"
             'dos' = $Dos
             'dat' = $Dat
-            'version' = "1.0"
+            'version' = "1.1"
+        }
+
+        $Redirect = Invoke-WebRequest -URI $Api -UseBasicParsing -Headers $Headers -MaximumRedirection 0
```
privateducky, line 41: I'm a bit worried messing with this too quickly; I'd recommend doing this server-side, in an "https://preludesecurity.com/upgrade" path

packplusplus (approved):
It doesn't really matter because the sed is going to sed, but the us1 bucket is in us-east-2. `prelude-account-us1-us-east-2.s3.amazonaws.com`

privateducky (approved):

PR #628, opened 2023-05-17
lowercase unit

makk94:


khyberspache (changes requested):
doesn't match the new names

PR #635, opened 2023-05-18
feat: listAdvisories and typing in JS SDK

stephanwampouille:
This PR adds the listAdvisories function for the JS SDK and returns a newly created Advisory object. As well, this PR updates the nomenclature for the applicable objects that were being used in "units" which are now "advisories"

### DIFF
--- a/javascript/sdk/lib/types.ts
+++ b/javascript/sdk/lib/types.ts
@@ -207,13 +207,6 @@ export interface TestData {
   mappings: string[];
 }
 
-export interface Unit {
-  id: string;
-  label: string;
-  definition: string;
-  description: string;
-}
-
 export interface AdvisoryUsage {
   count: number;
   failed: number;
@@ -221,10 +214,12 @@ export interface AdvisoryUsage {
 
 export interface AdvisoryInfo {
   id: string;
-  usage?: AdvisoryUsage;
+  usage: AdvisoryUsage;
 }
 
 export interface Advisory {
+  account: string;
+  created: string;
   id: string;
   name: string;
   link: string;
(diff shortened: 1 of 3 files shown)


javascript/sdk/lib/types.ts
```diff
@@ -214,15 +214,21 @@ export interface Unit {
...
 
-export interface UnitInfo {
-  unit: Unit;
-  usage?: UnitUsage;
+export interface AdvisoryInfo {
+  id: string;
+  usage?: AdvisoryUsage;
+}
+
+export interface Advisory {
```
makk94, line 227: advisories also have a created timestamp and an account (tho as of now the account will always be 'prelude')

javascript/sdk/lib/types.ts
```diff
@@ -214,15 +214,21 @@ export interface Unit {
...
   count: number;
   failed: number;
 }
 
-export interface UnitInfo {
-  unit: Unit;
-  usage?: UnitUsage;
+export interface AdvisoryInfo {
+  id: string;
+  usage?: AdvisoryUsage;
```
makk94, line 224: as of now, you'll only get advisories that have not-null/non-zero usage

javascript/sdk/lib/types.ts
```diff
@@ -207,22 +207,23 @@ export interface TestData {
...
-export interface UnitInfo {
-  unit: Unit;
-  usage?: UnitUsage;
+export interface AdvisoryInfo {
+  id: string;
+  usage: AdvisoryUsage;
+}
+
+export interface Advisory {
+  account: string;
```
makk94, line 221: do these fields need to match the json response exactly? (i forget how js works)

if so, it'll be `account_id`

makk94 (approved):

PR #637, opened 2023-05-18
Test alert param removed, advisory param added

ceci21:


stephanwampouille:
this can do into the 1.1.10 deploy which hasn't happened yet

kenrick (approved):

PR #639, opened 2023-05-18
fix: move getTest & download to detect controller

stephanwampouille:
@makk94 I see you updated the pathing but this PR updates where the code lives and properly moves them to the detect controller

### DIFF
--- a/javascript/sdk/lib/controllers/build.ts
+++ b/javascript/sdk/lib/controllers/build.ts
@@ -13,11 +13,12 @@ export default class BuildController {
     id: string,
     name: string,
     unit: string,
+    advisory: string,
     options: RequestOptions = {}
   ): Promise<void> {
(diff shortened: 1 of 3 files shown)


makk94:
ah my bad, didn't even check to see how you guys structured your controllers. thanks!

javascript/sdk/lib/controllers/build.ts
```diff
@@ -18,7 +17,7 @@ export default class BuildController {
   ): Promise<void> {
     await this.#client.requestWithAuth(`/build/tests/${id}`, {
       method: "POST",
-      body: JSON.stringify({ name, unit}),
+      body: JSON.stringify({ name, unit }),
```
makk94, line 20: create test will takes an "advisory" field now too (string)

kenrick (approved):

makk94 (approved):

PR #642, opened 2023-05-22
Update python and JS sdk tests

stephanwampouille:


khyberspache (approved):
All tests works for me against US2

PR #648, opened 2023-05-23
adding 'tests' activity view in JS SDK

stephanwampouille:


javascript/sdk/lib/types.ts
```diff
@@ -211,6 +211,16 @@ export interface TestData {
...
 }
 
+export interface TestUsage {
+  count: number;
+  failed: number;
+}
+
+export interface TestActivity {
+  id: string;
+  usafe: TestUsage;
```
makk94, line 221: usage

stephanwampouille: wow and thats why we review :P Thanks Mahina!

PR #659, opened 2023-05-27
upload via presigned url

makk94:
Bundle of PRs:
- https://github.com/preludeorg/service/pull/977
- https://github.com/preludeorg/backbone/pull/346
- https://github.com/preludeorg/compute/pull/110
- https://github.com/preludeorg/libraries/pull/659

Todo:
- update sdk tests

### DIFF
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -41,16 +41,17 @@ def create_test(controller, name, test, unit, advisory):
         template = template.replace('$NAME', name)
         template = template.replace('$UNIT', unit or '')
         template = template.replace('$CREATED', str(datetime.utcnow()))
-        
-        with Spinner():
-            controller.upload(test_id=test_id, filename=basename, data=template)
 
         test_dir = PurePath(test_id, basename)
         Path(test_id).mkdir(parents=True, exist_ok=True)
-        
+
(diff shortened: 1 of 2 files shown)


makk94:
Our backend creates presigned urls that only allow uploads up to 1MB. The error message from our CLI (if we propagate what comes back from s3) when you try to upload something bigger to that url is not good:
```
HTTPSConnectionPool(host='mahina-us-east-1.s3.amazonaws.com', port=443): Max retries exceeded with url: / (Caused by SSLError(SSLEOFError(8, 'EOF occurred in violation of protocol (_ssl.c:2396)')))
```
**Why?** httplib/urllib/requests libraries silently trash 100-continue responses, which the s3 presigned url uses on big uploads. I think this is the cause of that error. My solution in this PR is just to calculate content length ahead of the upload to s3 so we can error nicely

Note (mostly for myself if needed), using pycurl instead of requests does work/ give us expected nice error message from s3 (`Your proposed upload exceeds the maximum allowed size`)
```
        import pycurl
        import json
        c = pycurl.Curl()
        c.setopt(c.URL, res['url'])
        data = []
        for k, v in res['fields'].items():
            data.append((k, v))
        data.append(('file', (c.FORM_FILE, str(file))))
        c.setopt(c.HTTPPOST, data)
        c.perform()
        print(c.getinfo(pycurl.RESPONSE_CODE))
        c.close()
```

PR #667, opened 2023-06-01
Adding created ts to create test

privateducky:


stephanwampouille (commented):
I see "created" is optional in python SDK but required in JS. Why is this?

*edit my misunderstanding.

privateducky:
> I see "created" is optional in python SDK but required in JS. Why is this?

where are you looking, specifically?

stephanwampouille (approved):

privateducky:
@stephanwampouille plz do a follow up with the JS-SDK updates needed to match.

PR #673, opened 2023-06-05
update sdk for create/update test/endpoint

makk94:

PR #674, opened 2023-06-05
WIP update JS sdk for create/update test and endpoint

stephanwampouille:
parallel JS PR for #673 

As well as adding `state` to the `endpoints` response object

### DIFF
--- a/javascript/sdk/lib/controllers/build.ts
+++ b/javascript/sdk/lib/controllers/build.ts
@@ -18,7 +18,7 @@ export default class BuildController {
   ): Promise<Test> {
     const response = await this.#client.requestWithAuth(`/build/tests`, {
       method: "POST",
-      body: JSON.stringify({ name, unit, advisory, testId }),
+      body: JSON.stringify({ name, unit, advisory, id: testId }),
       ...options,
     });
 
--- a/javascript/sdk/lib/controllers/build.ts
+++ b/javascript/sdk/lib/controllers/build.ts
@@ -1,5 +1,5 @@
 import Client from "../client";
-import type { RequestOptions } from "../types";
+import type { RequestOptions, Test } from "../types";
 
 export default class BuildController {
   #client: Client;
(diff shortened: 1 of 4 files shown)


stephanwampouille:
holding in draft state until I update package version

javascript/sdk/lib/controllers/build.ts
```diff
@@ -10,17 +10,39 @@ export default class BuildController {
...
+    name?: string,
+    unit?: string,
+    advisory?: string,
+    options: RequestOptions = {}
+  ): Promise<Test> {
+    const response = await this.#client.requestWithAuth(
+      `/build/tests/${testId}`,
+      {
+        method: "POST",
+        body: JSON.stringify({ name, unit, advisory }),
```
makk94, line 40: so if user doesn't provide an advisory, does the body not contain the "advisory" key at all?

stephanwampouille: yep, is that not correct?

javascript/sdk/lib/controllers/build.ts
```diff
@@ -10,17 +10,39 @@ export default class BuildController {
...
     advisory?: string,
+    testId?: string,
     options: RequestOptions = {}
-  ): Promise<void> {
-    await this.#client.requestWithAuth(`/build/tests/${testId}`, {
+  ): Promise<Test> {
+    const response = await this.#client.requestWithAuth(`/build/tests`, {
       method: "POST",
-      body: JSON.stringify({ name, unit, advisory }),
+      body: JSON.stringify({ name, unit, advisory, testId }),
```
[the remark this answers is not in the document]
stephanwampouille: ah I see. fixed

javascript/sdk/lib/controllers/build.ts
stephanwampouille: it will be passed in as undefined I believe? @kenrick can you verify that statement

makk94: the updates work as "if the key is in the body, we'll update it's field. And if the key is not in the body, we leave the field as it is"

So we don't want keys to be passed through with default values, b/c if the user isn't explicitly setting the input param, they probably mean to leave it unchanged

makk94: so same for updateEndpoint - we don't want to default host/edr_id/tags to "" or all those will actually be updated to empty string in the BE

kenrick:
```js
JSON.stringify({ advisory: undefined })
```
will produce

```
"{}"
```

Undefined it not a valid JSON value. So the property would be omitted.

javascript/sdk/lib/controllers/detect.ts
```diff
@@ -27,27 +28,30 @@ export default class DetectController {
 
   /** Register (or re-register) an endpoint to your account */
   async registerEndpoint(
```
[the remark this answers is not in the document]
stephanwampouille:
```
  /** Register (or re-register) an endpoint to your account */
  async registerEndpoint(
    { host, serial_num, edr_id = "", tags }: RegisterEndpointParams,
    options: RequestOptions = {}
  ): Promise<string> {
    const response = await this.#client.requestWithAuth("/detect/endpoint", {
      method: "POST",
      body: JSON.stringify({ id: `${host}:${serial_num}:${edr_id}`, tags }),
      ...options,
    });

    return response.text();
  }
```

ok so I am defaulting edr_id to an empty string so that it will look like "HOST:SERIAL:" as you said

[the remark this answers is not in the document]
makk94: cool, all looks good to me!

only optional style thing is if you want to add question marks in register/update enpdoint to indicate which params are optional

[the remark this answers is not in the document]
stephanwampouille: so we are importing the typing from our types.ts file which you can see we are referring to it at `{ host, serial_num, edr_id = "", tags }: RegisterEndpointParams,` one of my goals is to remove the types file and directly place the typing using JS-doc but thats something for the future. It's in the works!

kenrick (approved):

makk94 (approved):

PR #675, opened 2023-06-06
Updating python and JS sdk tests

stephanwampouille:
- [x] waiting on BE change then should be good to go.
- [x] added new update endpoint/test tests

> all tests are passing except there are still inconsistencies with the probe timing out before both tests can run and describe-activity tests pass 100% of the time. I have seen it pass before on both python and JS today so I would consider it fine just inconsistent.

Update to this ^ @makk94 helped me debug, thank you!, but a solution to the issue that we were facing is that we were not giving compute enough time to handle the test attachments, so adding a 5s sleep did the trick. 
Screenshot 2023-06-06 at 3 54 49 PM
Screenshot 2023-06-06 at 3 53 09 PM

### DIFF
--- a/javascript/sdk/test/sdk.test.ts
+++ b/javascript/sdk/test/sdk.test.ts
@@ -194,18 +194,23 @@ describe("SDK Test", () => {
       activeTest = testId;
     });
 
-    it("download should return the same data as the upload", async () => {
-      const file = readFileSync(`${__dirname}/templates/template.go`, "utf8");
-      let data = file.toString();
-      data = data.replace("$ID", testId);
-      data = data.replace("$NAME", testName);
-      data = data.replace("$UNIT", testUnit);
-      data = data.replace("$CREATED", new Date().toISOString());
-      await service.build.upload(testId, templateName, data);
-      const download = await service.detect.download(testId, templateName);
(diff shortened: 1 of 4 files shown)


javascript/sdk/test/sdk.test.ts
```diff
@@ -114,7 +114,7 @@ describe("SDK Test", () => {
     });
 
     it("createTest should not throw an error", async () => {
-      await service.build.createTest(testId, testName, testUnit);
+      await service.build.createTest(testName, testUnit, undefined, testId);
```
makk94, line 117: is this 'undefined' really how JS sdks work? we can't name/default the params?

makk94:
any new update tests?

stephanwampouille:
> any new update tests?

yep let me do that now

javascript/sdk/test/sdk.test.ts
kenrick: @stephanwampouille i do prefer the CreateTestParams btw. You can see i used a similar pattern other functions

privateducky (approved):
@makk94 leaning on your review here; I don't see anything major.

makk94 (approved):

PR #676, opened 2023-06-06
add version to demo scripts

makk94:


privateducky (approved):
let's get @packplusplus to upgrade these in all environments

PR #695, opened 2023-06-21
CLI - Upload credentials file (feature proposal)

ceci21:
This PR adds the ability to configure a new account in the CLI by supplying a valid credentials (.ini) file exported from our UI using the optional `--file` parameter, or flag `-f`.

#### Usage:

```
prelude configure -f <file-path>
# or 
prelude configure --flag <file-path>
```

One difference between the current method of configuring the account (entering by line item) and this proposed one is that the configuration file requires the `profile` section header, and `hq`, `account`, and `token` values (throwing an error if keys or values are missing), whereas the line item method has default values set for the `profile` and `hq`.

## Demo / Screenshots
In the demo screenshots below, I aliased `cli` to my local structure while developing-- so `cli` corresponds to `prelude`.

#### Successful usage

Screenshot 2023-06-20 at 7 06 32 PM

#### Errors
Screenshot 2023-06-20 at 7 05 16 PM
Screenshot 2023-06-20 at 7 03 49 PM
Screenshot 2023-06-20 at 7 02 56 PM

### DIFF
--- a/python/cli/prelude_cli/views/configure.py
+++ b/python/cli/prelude_cli/views/configure.py
@@ -4,7 +4,7 @@ from configparser import ConfigParser, MissingSectionHeaderError
 
 
 @click.command()
-@click.option('-f', '--file', help="path to credentials file (.ini)", default=None, type=click.Path(exists=True))
+@click.option('-f', '--file', help="path to credentials (.ini) file", default=None, type=click.Path(exists=True))
 @click.pass_obj
 def configure(account, file):
     """ Configure your local keychain """
@@ -15,7 +15,7 @@ def configure(account, file):
 
     if file:
         config = ConfigParser()
-        
+
         try:
             config.read(file)
             profile = config.sections()[0]
--- a/python/cli/prelude_cli/views/configure.py
+++ b/python/cli/prelude_cli/views/configure.py
@@ -1,13 +1,47 @@
 import click
 
+from configparser import ConfigParser, MissingSectionHeaderError
+
 
 @click.command()
+@click.option('-f', '--file', help="path to credentials file (.ini)", default=None, type=click.Path(exists=True))
 @click.pass_obj
-def configure(account):
+def configure(account, file):
     """ Configure your local keychain """
-    profile = click.prompt('Enter the profile name', default=account.profile, show_default=True)
-    hq = click.prompt('Enter the Prelude API', default=account.hq, show_default=True)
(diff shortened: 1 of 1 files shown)


python/cli/prelude_cli/views/configure.py
```diff
@@ -1,13 +1,47 @@
...
-    profile = click.prompt('Enter the profile name', default=account.profile, show_default=True)
-    hq = click.prompt('Enter the Prelude API', default=account.hq, show_default=True)
-    account_id = click.prompt('Enter your account ID')
-    account_token = click.prompt('Enter your account token')
-    account.configure(account_id=account_id, token=account_token, hq=hq, profile=profile)
-    click.secho('Credentials saved', fg='green')
+    profile = account.profile
+    hq = account.hq
+    account_id = None
+    account_token = None
```
khyberspache, line 14: You don't need to declare these, python is smart enough to know that it will get set in both cases in your if statement below

python/cli/prelude_cli/views/configure.py
```diff
@@ -1,13 +1,47 @@
...
+
+    if file:
+        config = ConfigParser()
+
+        try:
+            config.read(file)
+            profile = config.sections()[0]
+            values = config[profile]
```
khyberspache, line 22: So it looks like we're taking the first entry from the file... what about that case where a keychain has multiple entries? Should we instead grab the "default" section and override the account.profile section? Or should that be configurable via additional flag? Thoughts?

khyberspache: Actually, the `Account` object has wrappers to handle the reading and importing of a keychain... 

Maybe we should refactor the Account class in Account.py to allow us to read in a keychain that's not the default keychain.

Here's the current function:
```python
def read_keychain_config(self, hq='https://api.preludesecurity.com', profile='default'):
        if not exists(self.keychain_location):
            head, _ = os.path.split(Path(self.keychain_location))
            Path(head).mkdir(parents=True, exist_ok=True)
            open(self.keychain_location, 'x').close()
            self.configure('', '', hq, profile)
        cfg = configparser.ConfigParser()
        cfg.read(self.keychain_location)
        return cfg
```

I think we can probably clean up this function to do something like:
```
try:
      valid_config = account.read_keychain_config(file_path=file, profile='default')
      account.configure(**valid_config)
 except:
     whoops bad config
```

khyberspache (changes requested):

PR #703, opened 2023-06-22
adding techniques field to create/update test

stephanwampouille:
All tests are passing.


makk94:
ideally a test somewhere would actually set the technique to make sure it works

PR #711, opened 2023-07-07
Implement JSDoc type handling on JS SDK

stephanwampouille:
This PR's focus is on handling types differently in the JS SDK and converts the primary usage from typescript to jsdoc. This will eliminate the need for building the typescript files to javascript and reduce the number of minor publishes we have to complete to update type changes. The types will now be handled both here and platform so small changes can be made independent from each other and serve as a quicker way to make updates on platform.


javascript/sdk/lib/client.js
```diff
@@ -0,0 +1,197 @@
...
+  return headers.reduce((acc, h) => ({ ...acc, ...h }), {});
+}
+
+/**
+ * Custom error class for API errors.
+ *
+ * @class
+ * @extends Error
+ */
+class APIError extends Error {
```
kenrick, line 40: Can you also export this so I check for this error in the UI

PR #720, opened 2023-07-17
Update partner endpoints to use new control codes

stephanwampouille:
This PR updates the /partner endpoints to take in a `partner-code` which is an enum value. For example `endpoint/deploy/1`

Thoughts on naming convention @makk94 ? Not sure if I should switch between partner and control

### DIFF
--- a/javascript/sdk/lib/controllers/partner.ts
+++ b/javascript/sdk/lib/controllers/partner.ts
@@ -1,6 +1,7 @@
 import Client from "../client";
 import {
   AttachPartnerParams,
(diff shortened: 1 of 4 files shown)


privateducky (approved):
I would merge, but this is still in draft

makk94:
> Thoughts on naming convention @makk94 ? Not sure if I should switch between partner and control

no strong opinion from me, using `partner` in the sdk seems fine to me

stephanwampouille:
> I would merge, but this is still in draft

Good to go now

PR #722, opened 2023-07-17
CLI - Add partner code params to partner attach, detach, deploy, endpoints

ceci21:


python/cli/prelude_cli/views/partner.py
```diff
@@ -14,52 +14,52 @@ def partner(ctx):
 
 
 @partner.command('attach')
-@click.argument('name')
+@click.argument('partner-code')
```
privateducky, line 17: two questions:
1. is partner-code the best name?
2. how could a user become aware of the options?

ceci21: I'll add help text and elaborate. Might be a good time to throw in the enum as well.

privateducky (approved):

PR #730, opened 2023-07-25
ExitCode transform - not a class method

makk94:


stephanwampouille:
Could you explain this to me? The difference in how we set it here and how its used in the BE?

privateducky (approved):

PR #731, opened 2023-07-27
update endpoint responses for JS and Python

stephanwampouille:
This PR updates the responses of multiple endpoints (primarily for JS SDK as we comply with types) following the responses listed [here](https://preludeorg.slack.com/docs/T7GV7S3B4/F05JXGQ2W3C). To recap some changes:

- Adding the test attachments to the test object that gets returned from getTest, createTest, and updateTest
- Update multiple endpoints, primarily those related to deletes, to return a {status:true} response 
- Return the attachment on upload-attachment
- Add type for AuditLog and update endpoint to return that type
- Smaller updates of minor significance on other endpoints and small refactors of code
- Updated tests to work with new responses

Checklist:
- [x] javascript
- [x] javascript tests
- [x] python
- [x] python tests

### DIFF
--- a/javascript/sdk/lib/types.ts
+++ b/javascript/sdk/lib/types.ts
@@ -435,8 +435,4 @@ export interface AuditLog {
   timestamp: string;
 }
 
-export interface AuditLogValues {
-  name: string;
-  unit: string;
-  id: string;
-}
+export type AuditLogValues = Record<string, unknown>;
--- a/javascript/sdk/lib/types.ts
+++ b/javascript/sdk/lib/types.ts
@@ -413,7 +413,7 @@ export interface DeployParams {
 }
 
 export interface DeployedEndpoints {
-  id: string;
+  id: ControlCode;
   host_ids: string[];
 }
 
@@ -431,7 +431,7 @@ export interface AuditLog {
   account_id: string;
   user_id: string;
   values: AuditLogValues;
-  status: ExitCode;
+  status: string;
   timestamp: string;
 }
 
--- a/javascript/sdk/lib/types.ts
+++ b/javascript/sdk/lib/types.ts
@@ -13,6 +13,10 @@ export interface Credentials {
 
 export type RequestOptions = Omit<RequestInit, "method" | "body">;
 
+export interface StatusResponse {
+  status: true;
+}
+
 export interface Test {
   account_id: string;
   id: string;
@@ -20,6 +24,15 @@ export interface Test {
   unit: string;
   techniques: string[];
   advisory: string;
+}
+
+export type AttachedTest = Test & {
+  attachments: string[];
+};
+
+export interface UploadedAttachment {
+  id: string;
+  filename: string;
 }
 
 export interface User {
@@ -36,10 +49,10 @@ export interface Control {
 export interface Account {
   account_id: string;
   whoami: string;
-  controls: Control[];
   users: User[];
-  mode: Mode;
   queue: Queue[];
+  controls: Control[];
+  mode: Mode;
   company: string;
 }
 
@@ -119,6 +132,10 @@ export interface EnableTest {
   tags?: string;
 }
 
+export interface EnabledTest {
+  id: string;
+}
+
 export interface DisableTest {
   test: string;
   tags: string;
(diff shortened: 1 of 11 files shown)


kenrick (approved):

javascript/sdk/lib/types.ts
```diff
@@ -399,3 +425,18 @@ export type PartnerEndpoints = Record<
...
+export interface AuditLog {
+  name: string;
+  account_id: string;
+  user_id: string;
+  values: AuditLogValues;
+  status: ExitCode;
+  timestamp: string;
+}
+
+export interface AuditLogValues {
```
makk94, line 438: Is there the idea of "any" or "blob" in js? `values` holds different fields for different events and ideally is a blackbox dictionary (ex. create_user will probably have permission and handle, upload_attachment will have test_id and filename, etc)

javascript/sdk/lib/types.ts
```diff
@@ -378,6 +394,11 @@ export interface AttachPartnerParams {
   secret?: string;
 }
 
+export interface AttachedPartner {
```
makk94, line 397: I'm a little surprised this didn't already exist? This has been the response for a little while now

javascript/sdk/lib/types.ts
```diff
@@ -399,3 +425,18 @@ export type PartnerEndpoints = Record<
     state: string;
   }
 >;
+
+export interface AuditLog {
+  name: string;
```
makk94, line 430: just merged a PR (not yet deployed) that'll make this field named `event` instead of `name`

stephanwampouille: thank you. updated!

javascript/sdk/lib/types.ts
stephanwampouille: @kenrick So im thinking we could make the "values" in AuditLog the type of "any" or perhaps just an object?

kenrick: type AuditLogValues = Record<string, unknown>

javascript/sdk/lib/types.ts
kenrick: probably was, the UI didn't use it so the type wasn't update when it was changed.

javascript/sdk/lib/types.ts
stephanwampouille: ah that would make more sense

javascript/sdk/lib/types.ts
```diff
@@ -391,6 +412,11 @@ export interface DeployParams {
   hostIds: string[];
 }
 
+export interface DeployedEndpoints {
+  id: string;
```
makk94, line 416: fyi - this id is the enum value of the Partner (1 for crowdstrike, etc). Know you added that ControlCode enum somewhere, not sure if you want to tie in here in some way?

javascript/sdk/lib/types.ts
```diff
@@ -399,3 +425,14 @@ export type PartnerEndpoints = Record<
     state: string;
   }
 >;
+
+export interface AuditLog {
+  event: string;
+  account_id: string;
+  user_id: string;
+  values: AuditLogValues;
+  status: ExitCode;
```
makk94, line 434: also just wanna be clear that this status value is the http status (not a probe ExitCode status). It'll be a string, but it'll be 200, 400, 500, etc

python/sdk/prelude_sdk/controllers/detect_controller.py
```diff
@@ -135,7 +135,7 @@ def download(self, test_id, filename):
             timeout=10
         )
         if res.status_code == 200:
-            return res.content
+            return res.text
```
makk94, line 138: wonder if this should stay content - some attachments might be binary

stephanwampouille: yes I updated it back after I pinged you yesterday!

javascript/sdk/lib/types.ts
stephanwampouille: updated!

javascript/sdk/lib/types.ts
stephanwampouille: updated!

makk94 (approved):

kenrick (approved):

privateducky (approved):

PR #749, opened 2023-08-09
set LC_ALL; fix nocturnal on mac

makk94:


khyberspache (approved):
Need to set locale to basic C types only

PR #753, opened 2023-08-11
det-605: add partner_block sdk/cli

makk94:
PR bundle:
- https://github.com/preludeorg/partners/pull/51
- https://github.com/preludeorg/service/pull/1219
- https://github.com/preludeorg/libraries/pull/753
- https://github.com/preludeorg/backbone/pull/470

### DIFF
--- a/python/cli/prelude_cli/views/partner.py
+++ b/python/cli/prelude_cli/views/partner.py
@@ -42,9 +42,22 @@ def detach_partner(controller, partner):
     print_json(data=data)
 
 
+@partner.command('block')
+@click.argument('partner',
+              type=click.Choice([c.name for c in Control if c.is_edr], case_sensitive=False))
+@click.option('-t', '--test_id', required=True, help='a test to block')
+@click.pass_obj
+@handle_api_error
+def partner_deploy(controller, partner, test_id):
+    """ Report to a partner to block a test """
+    with Spinner(description='Reporting test to partner'):
+        data = controller.block(partner_code=Control[partner.upper()].value, test_id=test_id)
+    print_json(data=data)
+
+
 @partner.command('endpoints')
 @click.argument('partner',
-              type=click.Choice([c.name for c in Control if c != Control.INVALID], case_sensitive=False))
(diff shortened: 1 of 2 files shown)


python/sdk/prelude_sdk/controllers/partner_controller.py
```diff
@@ -33,10 +33,24 @@ def detach(self, partner_code: int):
         if res.status_code == 200:
             return res.json()
         raise Exception(res.text)
+
+    @verify_credentials
+    def block(self, partner_code: int, test_id: str):
+        """ Report to a partner to block a test """
+        params = dict(test_id=test_id)
```
stephanwampouille, line 40: Why are we sending the test_id rather than the something unique to the activity like it's id? On the front end, per our convo yesterday it sounded like we wanted to send in a single activity rather than a test which would be a culmination of activities if they are ran multiple times. I could be mistaken though

python/sdk/prelude_sdk/controllers/partner_controller.py
```diff
@@ -33,10 +33,24 @@ def detach(self, partner_code: int):
...
+        res = requests.post(
+            f'{self.account.hq}/partner/block/{partner_code}',
+            headers=self.account.headers,
+            json=params,
+            timeout=30
+        )
+        if res.status_code == 200:
+            return res.json()
```
stephanwampouille, line 48: also what does this return?

python/cli/prelude_cli/views/partner.py
```diff
@@ -60,7 +73,7 @@ def partner_endpoints(controller, partner, platform, hostname, offset):
 @partner.command('deploy')
 @click.confirmation_option(prompt='Are you sure?')
 @click.argument('partner',
```
privateducky, line 75: great change/eye

privateducky (approved):

python/sdk/prelude_sdk/controllers/partner_controller.py
privateducky: @stephanwampouille we'll want this to go on the JS-SDK as well.

PR #765, opened 2023-08-17
Make is_edr intrinsic to the backend implementation

khyberspache:


makk94:
is having the is_edr property defined on every  partner  better than using `partner.id.is_edr` in bb code?

makk94:
also, if were going this route, i didn't see `is_siem` migrated over to the partners

privateducky:
> BB code shouldn't have explicit partner.id.is_edr calls, if so we should adjust.
> 
> Is_siem isn't used anywhere i could find

is_siem is only used in probe_tasks

makk94:
> BB code shouldn't have explicit partner.id.is_edr calls, if so we should adjust

or service code, or wherever it is.

Seems like much more code to define a property on each partner impl instead of have one mapping of "type" on enum

khyberspache:
I updated is_siem to be a negated is_edr in service.

On the other point: We're removing a step.

Right now it's:
1. Dict of control
2. Check if control is_edr in Enum
3. Construct partner object
4. Handle partner object actions


We are just saying:

1. Dict of control
2. Construct partner object
3. Handle partner object actions

PR #766, opened 2023-08-17
Add a Category enum for account categories

khyberspache:


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -29,6 +29,16 @@ def _missing_(cls, value):
         return Mode.MANUAL
 
 
+class Category(Enum):
+    FREE = 0
+    INTNERAL = 1
```
makk94, line 34: spelling

PR #767, opened 2023-08-17
remove control & EDR ID from update/register endpoint

stephanwampouille:
I want to double check that that id in the registerEndpoint is correct. Not sure if the extra `:` was needed to be kept at the end

### DIFF
--- a/javascript/sdk/lib/controllers/detect.ts
+++ b/javascript/sdk/lib/controllers/detect.ts
@@ -5,7 +5,6 @@ import {
   Advisory,
   AdvisoryInfo,
   AttachedTest,
-  ControlCodes,
   DayResult,
   DisableTest,
   EnableTest,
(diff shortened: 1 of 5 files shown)


javascript/sdk/lib/controllers/detect.ts
```diff
@@ -56,13 +49,7 @@ export default class DetectController {
...
   /** Update an endpoint in your account */
   async updateEndpoint(
-    {
-      endpoint_id,
-      host,
-      edr_id,
-      control = ControlCodes.INVALID,
-      tags,
-    }: UpdateEndpointParams,
+    { endpoint_id, host, tags }: UpdateEndpointParams,
```
privateducky, line 52: Need python change too; we don't want the last :

privateducky (changes requested):

privateducky (approved):

PR #779, opened 2023-08-23
Chore: remove unused run code & hard code sleep

privateducky:
Sleep is there only for debug purposes, which I will be solving separately this week.


stephanwampouille:
Do you want me to wait to make a JS equivalent PR?

privateducky:
> Do you want me to wait to make a JS equivalent PR?

Yessir; just note that the UI will need a fallback for 1 day, if the DEBUG code is removed from the JS-SDK.

stephanwampouille:
@kenrick Note that we will need to gracefully handle a fallback. Do we have any thoughts on a general verbiage for it?

stephanwampouille (approved):

PR #783, opened 2023-08-24
Chore: use the test ID as the filename when writing to disk

privateducky:
- Make PRELUDE_DIR not configurable
- Remove print statement
- Split curl into 2 calls

### DIFF
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -31,7 +31,7 @@ while ($true) {
     try {
         $VstDir = New-Item -ItemType Directory -Path ".vst" -Force
         $Headers = @{
-            'token' = "c9bcc556d45abbc4b01113be65bbf43a"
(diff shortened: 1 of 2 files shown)


khyberspache (approved):

khyberspache:
We will need to test these with the installers, but should be g2g

shell/probe/raindrop.ps1
```diff
@@ -30,23 +29,22 @@ $Dat = ""
 
 while ($true) {
     try {
-        $Vst = New-Item -Path "$Dir\$(New-Guid).exe" -Force
+        $VstDir = New-Item -ItemType Directory -Path ".vst" -Force
         $Headers = @{
-            'token' = FromEnv "PRELUDE_TOKEN"
+            'token' = "c9bcc556d45abbc4b01113be65bbf43a"
```
privateducky, line 34: Hard code

privateducky (commented):
Hard coded token

PR #786, opened 2023-08-25
revert change to env var

privateducky:


shell/probe/raindrop.ps1
```diff
@@ -29,7 +30,6 @@ $Dat = ""
 
 while ($true) {
     try {
-        $VstDir = New-Item -ItemType Directory -Path ".vst" -Force
```
khyberspache, line 32: Keep this, just pass the full path here, leave everything else the same

khyberspache (changes requested):
One adjustment

shell/probe/raindrop.ps1
privateducky: ha, got this just as i did.. can you check if my change should work?

khyberspache: Like use $Dir as the argument to this line, keep everything else the same

privateducky: did i do it...?

khyberspache: No lol you need to capture this to a variable, just revert all your changes and add the one line at the top to pull the directory from the environment

privateducky: oh - i THINK i see... like this?

privateducky: i totally got it, didn't i?

khyberspache: You did :)

PR #792, opened 2023-08-28
Adding Endpoint Print func

gerbsec:


bfuzzy1 (approved):
@gerbsec can you update all the tests in tests/ master to use this new print function please?

Approving since we tested this together.

bfuzzy1:
@privateducky would like your approval on this change as well. Also we should probably hold back on merging this until the swap for real test ids goes in this week.

go/tests/endpoint/endpoint.go
```diff
@@ -44,7 +44,7 @@ func Stop(code int) {
 func Print(print string) {
 	filename := filepath.Base(os.Args[0])
 	name := strings.TrimSuffix(filename, filepath.Ext(filename))
-	Print(fmt.Sprintf("[%s] %s\n", name, print))
+	fmt.Sprintf("[%s] %s\n", name, print)
```
privateducky: check out ``fmt.Sprintf("[%s] %v", name, print)``

go/tests/endpoint/endpoint.go
```diff
@@ -23,7 +23,7 @@ func Start(test fn, clean ...fn) {
 		cleanup = clean[0]
 	}
 
-	Print("Starting test")
+	Print(fmt.Sprintf("Starting test at: %s", time.Now().UTC().Format("2006-01-02T15:04:05Z")))
```
privateducky, line 26: we should do local time

bfuzzy1: I fix.

bfuzzy1: fixed!

privateducky (approved):

PR #796, opened 2023-08-29
[DET-564] Password based authentication

khyberspache:


stephanwampouille:
I see the similarity as you mentioned to how AWS does it

khyberspache:
YARP

PR #797, opened 2023-08-29
chore: add dos to probe object

stephanwampouille:
This PR adds the dos property to the probe object and assigns the type Platform to it, for reference Platform is:

export type Platform =
  | "darwin-arm64"
  | "darwin-x86_64"
  | "linux-x86_64"
  | "linux-arm64"
  | "windows-x86_64"
  | "windows-arm64"
  | "unknown";
  
  @kenrick for your awareness


privateducky (approved):
great!

PR #803, opened 2023-08-31
[DET-706] Add an AUTO permission for webhook auth and CLI stuff for generating webhook

khyberspache:


javascript/sdk/lib/controllers/partner.ts
```diff
@@ -112,4 +112,20 @@ export default class PartnerController {
...
+  ): Promise<StatusResponse> {
+    const response = await this.#client.requestWithAuth(
+      `/partner/suppress/${partnerCode}`,
+      {
+        method: "GET",
+        ...options,
+      }
+    );
+
+    return response.json();
```
stephanwampouille, line 129: this might need to be awaited?

```suggestion
    return await response.json();
```

stephanwampouille (changes requested):

javascript/sdk/lib/controllers/partner.ts
khyberspache: Make edits as necessary

stephanwampouille: updated btw

python/cli/prelude_cli/views/partner.py
```diff
@@ -81,3 +81,14 @@ def partner_deploy(controller, partner, host_ids):
     with Spinner(description='Deploying probes to hosts'):
         data = controller.deploy(partner_code=Control[partner.upper()].value, host_ids=host_ids)
     print_json(data=data)
+
+
+@partner.command('webhook-auth')
+@click.argument('partner', type=click.Choice([c.name for c in Control if c != Control.INVALID], case_sensitive=False))
```
privateducky, line 87: should this be restricted to EDR controls instead? or do you think doesn't matter/too niche?

python/cli/prelude_cli/views/partner.py
```diff
@@ -81,3 +81,14 @@ def partner_deploy(controller, partner, host_ids):
...
+@click.pass_obj
+@handle_api_error
+def feed_auth(controller, partner):
+    """ Get feed authentication information for a partner """
```
privateducky, line 91: can you add a better comment, more user-friendly?

privateducky: same with the command; i'd use something like ``webook-create`` or ``webhook-generate``. it's more descriptive of what you're doing in totality.. then i'd probably put a print line that points to our docs on how to finish the work/next steps.

privateducky (changes requested):
couple small items

python/cli/prelude_cli/views/partner.py
khyberspache: Yeah that's good point - I think it's worth adding the check. I'll do that

python/cli/prelude_cli/views/partner.py
khyberspache: I think webhook-generate is a good option. I'll shift to that

privateducky (approved):

PR #807, opened 2023-08-31
Update ca to "true" location

privateducky:
I'm told our pipeline will need an update.


khyberspache:
Uhhh us 1 is not in east 1 dawg

PR #810, opened 2023-09-01
chore: adjusting run codes

privateducky:


stephanwampouille (approved):
Will this require a DB migration?

privateducky:
> Will this require a DB migration?

No; because the run code hasn't been released yet.

PR #816, opened 2023-09-05
chore: update activity views and fix tests (JS SDK 1.4.9)

stephanwampouille:
This PR updates the typing for the JS DescribeActivity to remove the `tests` view and update the `social` and `day` responses.

social returns:
`{
  account: { unprotected: number; volume: number };
  social: { unprotected: number; volume: number };
}`

and day returns: 
`{
  unprotected: number;
  volume: string;
}`

This PR also fixes the sdk tests to account for accounts being on autopilot mode by default

### DIFF
--- a/javascript/sdk/lib/types.ts
+++ b/javascript/sdk/lib/types.ts
@@ -100,7 +100,7 @@ export const RunCodes = {
   FRIDAY: 14,
   SATURDAY: 15,
   SUNDAY: 16,
-  MONTH_1: 20
+  MONTH_1: 20,
 } as const;
 
 export type RunCode = (typeof RunCodes)[keyof typeof RunCodes];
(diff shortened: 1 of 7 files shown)


stephanwampouille:
I also noticed we are returning a {status: true} for updateAccount now, which seems new, so I updated that as well

javascript/sdk/lib/types.ts
```diff
@@ -275,10 +273,9 @@ export interface ActivityQuery {
   control?: ControlCode;
 }
 
-export interface DayResult {
-  count: number;
-  datetime: string;
-  failed: number;
+export interface DayActivity {
```
makk94, line 276: it looks like there's still a `datetime` field 
`days.append(dict(datetime=h[0], unprotected=h[1], volume=h[2]))`

stephanwampouille: re-added :)

makk94 (approved):

PR #819, opened 2023-09-06
chore: revive advisories and tests view

stephanwampouille:
This PR brings back the advisories and tests views and adds in the optional field of "impersonate" for the activity query

### DIFF
--- a/javascript/sdk/lib/types.ts
+++ b/javascript/sdk/lib/types.ts
@@ -243,10 +243,10 @@ export interface Activity {
   control: ControlCode | null;
 }
 
-export interface TestUsage {
-  count: number;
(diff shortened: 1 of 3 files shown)


javascript/sdk/lib/types.ts
```diff
@@ -271,6 +277,7 @@ export interface ActivityQuery {
   statuses?: string;
   tags?: string;
   control?: ControlCode;
+  impersonate?: string;
```
makk94, line 280: i'd expect some deletion code somewhere since impersonate is no longer in the headers? not a bug/ just possible dead code somewhere

makk94 (approved):

PR #830, opened 2023-09-12
update activity types and remove tags

stephanwampouille:
This PR does the following:

- Remove tags from "insights", "probes", and "logs" views
- Add impersonate to the query
- Remove deprecated "observed" property from "logs"
- Remove null from control for endpoints and "probes" view responses
- Add null option for dos on endpoints response

### DIFF
--- a/javascript/sdk/lib/controllers/detect.ts
+++ b/javascript/sdk/lib/controllers/detect.ts
@@ -168,6 +168,7 @@ export default class DetectController {
     if (query.result_id) searchParams.set("result_id", query.result_id);
     if (query.dos) searchParams.set("dos", query.dos);
     if (query.control) searchParams.set("control", query.control.toString());
+    if (query.impersonate) searchParams.set("impersonate", query.impersonate);
 
     const response = await this.#client.requestWithAuth(
       `/detect/activity?${searchParams.toString()}`,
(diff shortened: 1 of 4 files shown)


javascript/sdk/lib/controllers/detect.ts
```diff
@@ -168,6 +168,7 @@ export default class DetectController {
     if (query.result_id) searchParams.set("result_id", query.result_id);
     if (query.dos) searchParams.set("dos", query.dos);
     if (query.control) searchParams.set("control", query.control.toString());
+    if (query.impersonate) searchParams.set("impersonate", query.impersonate);
```
kenrick, line 171: 👍

kenrick (approved):

javascript/sdk/lib/controllers/detect.ts
```diff
@@ -168,6 +168,7 @@ export default class DetectController {
     if (query.result_id) searchParams.set("result_id", query.result_id);
```
makk94, line 168: activity filters in the BE are:
- start
- finish
- tests
- dos
- endpoints
- statuses
- control

makk94 (approved):

javascript/sdk/lib/controllers/detect.ts
stephanwampouille: no impersonate?

stephanwampouille: It should be there since we are using that for social activity for metrics, tests, and advisories views

makk94: oh yeah, that too. wasn't really considering it a filter, but suppose it is 👍

privateducky (approved):

PR #831, opened 2023-09-13
Update Demo links

bfuzzy1:
Updated links to point at the advisory page since our old public test repo has been deleted.


privateducky (approved):
@packplusplus please push -- only these probes -- to us1/us2.

PR #832, opened 2023-09-13
add power and default views (JS SDK 1.4.14)

stephanwampouille:
Adds new power and default views. Adds powerActivity typing for the power view response. In that response, we say that policy, version, dos, and control are all optional properties and may not be returned from the BE

Adjusts for better checks on tests

### DIFF
--- a/javascript/sdk/lib/types.ts
+++ b/javascript/sdk/lib/types.ts
@@ -248,7 +248,7 @@ export interface Activity {
 
 export interface TestsActivity {
   id: string;
-  unprotected: number;
+  protected: number;
   volume: number;
 }
 
@@ -261,7 +261,7 @@ export interface Advisory {
 
 export interface AdvisoriesActivity {
   id: string;
-  unprotected: number;
+  protected: number;
   volume: number;
 }
 
@@ -311,7 +311,7 @@ export interface Recommendation {
 }
 
 export interface PowerActivity {
-  unprotected: number;
+  protected: number;
   dos?: Platform;
   control?: ControlCode;
   policy?: string;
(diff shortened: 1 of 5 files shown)


privateducky (approved):

privateducky:
@makk94 if you agree, merge

kenrick (approved):

javascript/sdk/lib/types.ts
```diff
@@ -244,7 +248,7 @@ export interface Activity {
 
 export interface TestsActivity {
   id: string;
-  unprotected: number;
+  protected: number;
   volume: number;
```
makk94, line 252: volume looks removed on the BE

javascript/sdk/lib/types.ts
```diff
@@ -257,7 +261,7 @@ export interface Advisory {
 
 export interface AdvisoriesActivity {
   id: string;
-  unprotected: number;
+  protected: number;
   volume: number;
```
makk94, line 265: same here. volume removed

javascript/sdk/lib/types.ts
```diff
@@ -289,6 +293,7 @@ export interface ProbeActivity {
   endpoint_id: string;
   state: "PROTECTED" | "UNPROTECTED" | "ERROR";
   control: ControlCode;
+  status?: number;
```
makk94, line 296: is this something new you were promised? I dont see this on the BE

stephanwampouille: Yeah I was told that there is going to be a status property on probe activity for use on the hosts table

stephanwampouille: Might not be there yet, but was adding types so that Kenrick can finish the table and then once its in the BE we can remove the optional field part of it

makk94 (approved):

privateducky (approved):

PR #834, opened 2023-09-13
Add Results struct and print table method

kenrick:


go/probe/debug/main.go
```diff
@@ -117,12 +131,13 @@ func loop(testID string, dat string) {
 
 				if cmd.ProcessState != nil {
 					code := cmd.ProcessState.ExitCode()
+					pr.Evaluations = append(pr.Evaluations, Evaluation{Policy: "policy1", Code: code})
```
matterpreter, line 134: We'll need to replace the hardcoded policy name on this line when we start iterating over the collection of policies we want to evaluate.

PR #847, opened 2023-09-20
det-772: add enum for audit events

makk94:


privateducky (approved):
Will merge Pending other PRs

PR #861, opened 2023-09-26
det-772 [a]: pass enum strings to API

makk94:
* have sdk take enum params instead of ints
* sdk then sends string names instead of ints to API

* **Added EnumMeta to our codes:** By default a non-existent enum string  will throw a KeyError. The EnumMeta `__getitem__` will catch that error and redirect to the normal enum constructor.
   * so anything using ints will continue to work (backwards compatible woo) (ex. `RunCode[2]` is `RunCode.WEEKLY`)
   * and non-existent string will  fall through to the `__missing__` method (ex. `RunCode['mahinas-birthday']` will be `RunCode.DAILY`)

with https://github.com/preludeorg/service/pull/1448

### DIFF
--- a/python/sdk/prelude_sdk/models/codes.py
+++ b/python/sdk/prelude_sdk/models/codes.py
@@ -6,7 +6,7 @@ class MissingItem(EnumMeta):
         try:
             return super().__getitem__(name.upper())
         except KeyError:
-            return cls(None)
+            return cls(name)
 
 
 class RunCode(Enum, metaclass=MissingItem):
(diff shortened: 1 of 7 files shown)


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -1,7 +1,15 @@
-from enum import Enum
+from enum import Enum, EnumMeta
 
 
-class RunCode(Enum):
+class MissingItem(EnumMeta):
```
privateducky, line 4: Nice

privateducky (approved):

privateducky:
@makk94 quick conflict on this one

PR #886, opened 2023-10-05
Adding random command selection function to Endpoint

matterpreter:
An increasing number of our tests rely on the following pattern to execute a randomly selected command on each test execution.

```go
func check() {
    commands := [][]string{
        {"cmd.exe", "/C", "foo.exe bar"},
        {"cmd.exe", "/C", "c:\Windows\System32\bar.exe 123"},
        { "powershell.exe", "-c", "Get-Foobar" },
    }
    index := rand.Intn(len(commands))
    command := commands[randomIndex]
    Endpoint.Shell(command)
}
```

This is a valuable feature of our tests as it lets us extend a give test to exercise multiple variations of the same operation. To reduce code reuse and standardize our implementation (no more `math/rand` vs `crypto/rand`), this new Endpoint function abstracts command selection away from the developer and leverages the exists, tested `Shell` function to execute the randomly chosen command.

### DIFF
--- a/go/tests/endpoint/endpoint.go
+++ b/go/tests/endpoint/endpoint.go
@@ -143,6 +143,20 @@ func Shell(args []string) (string, error) {
 	return string(stdout), nil
 }
 
+func ExecuteRandomCommand(commands [][]string) (string, error) {
+	var command []string
+	if len(commands) == 0 {
+		return "", fmt.Errorf("command slice is empty")
+	} else if len(commands) == 1 {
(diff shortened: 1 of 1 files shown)


bfuzzy1:
This is PR'd over in our test repo already and approved there: https://github.com/preludeorg/tests/pull/49

bfuzzy1 (approved):

bfuzzy1:
merging.

PR #887, opened 2023-10-05
chore: add type generated webhook

kenrick:


javascript/sdk/lib/types.ts
```diff
@@ -417,3 +417,9 @@ export interface BlockResponse {
   already_reported?: boolean;
   ioc_id?: string;
 }
+
+export interface GeneratedWebhook {
```
makk94, line 421: s1 returns that Header dict too

javascript/sdk/lib/controllers/partner.ts
```diff
@@ -115,7 +116,7 @@ export default class PartnerController {
   async webhookGenerate(
     partnerCode: ControlCode,
     options: RequestOptions = {}
-  ): Promise<StatusResponse> {
+  ): Promise<GeneratedWebhook> {
     const response = await this.#client.requestWithAuth(
       `/partner/suppress/${partnerCode}`,
```
makk94, line 121: also we moved to enum names and I think ControlCode is still the int value?

javascript/sdk/lib/types.ts
kenrick: @makk94 fixed :D

makk94: does it need a `?` or something since crowdstrike doesnt return it?

javascript/sdk/lib/controllers/partner.ts
kenrick: fixed!

packplusplus:
@kenrick we still need this?

PR #890, opened 2023-10-20
Add policy_name (JS SDK 1.5.6)

stephanwampouille:
LMK if it cant be undefined @makk94


makk94:
Do you need to add it to the partner endpoints response too? Or just don't care about that one

PR #891, opened 2023-10-23
DET-832: remove guard so we download files without suffixes

packplusplus:
About the only issue I could see here, would be if we started supported directories in tests. But I really don't want to think about that right now, and this is a demonstrated problem on an existing us2 test. Also, if we ever want to push 0fe330f2-7a9c-48c0-a268-5600b4a15ea7 to us1, we won't get that file.

### DIFF
--- a/python/cli/prelude_cli/views/detect.py
+++ b/python/cli/prelude_cli/views/detect.py
@@ -80,10 +80,9 @@ def download(controller, test):
         attachments = controller.get_test(test_id=test).get('attachments')
 
         for attach in attachments:
-            if Path(attach).suffix:
-                code = controller.download(test_id=test, filename=attach)
-                with open(PurePath(test, attach), 'wb') as f:
-                    f.write(code)
+            code = controller.download(test_id=test, filename=attach)
+            with open(PurePath(test, attach), 'wb') as f:
+                f.write(code)
 
 
 @detect.command('enable-test')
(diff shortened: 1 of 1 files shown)


makk94:
I comment out that `if has suffix` line all the time locally lol

stephanwampouille:
Ok so I get the if statement that checks if there is a suffix to the attachment but maybe I need a bit more explanation on why this isn't necessary?

packplusplus:
Oh! There was a requirement a while ago that all files needed suffixes. And then we dropped that requirement. There's nothing that prevents files from being uploaded / downloaded without a suffix in the api, just in the cli for downloads...

* backbone is fine uploading no suffix: https://github.com/preludeorg/backbone/blob/a4c19a0af1a9c3ce40f445001ed7b130645fee59/backbone/modules/impl/test.py#L52
* service is fine too https://github.com/preludeorg/service/blob/2b753a1fe96dac2ce6b68bb029ed5e9c628db97e/functions/upload_attachment/app.py#L10
* cli fine with uploading anything that's a file https://github.com/preludeorg/libraries/blob/e2bb76e328e50b49a8e5530fcd6330d7773587ff/python/cli/prelude_cli/views/build.py#L98-L130

stephanwampouille:
Ah thank you for the explanation, makes a lot more sense now! Good catch on the code, I approve!

stephanwampouille (approved):

makk94:
Do you wanna update `clone` too?

packplusplus:
@makk94 good catch! Updated the PR.

makk94 (approved):

PR #899, opened 2023-11-06
Feature: Add ResponseData struct to Network Package

bfuzzy1:
This change adds the ResponseData struct to the Network package so we have a cleaner way of capturing that output across tests.


bfuzzy1:
This has been approved and merged in the test repo: https://github.com/preludeorg/tests/pull/83

gerbsec (commented):
Looks good! But GET doesn't seem to be returning anything?

gerbsec (approved):
Very good!

PR #906, opened 2023-11-21
fix: handle bad exit code errors (like empty ones)

makk94:


packplusplus (changes requested):
Unless we can get the info from the logging decorator in there, I'm less apt to want to merge this.

makk94:
I think the fact that this errors out the probe from continuing to run the next test is more important than losing our little bit of logging

I'd prefer we have the xray stuff up instead of one off logs everywhere, but if it's a blocker for you, I'll add an `if exitCode == MISSING: then log (dos,dat,whatever)` to probe_tasks

PR #909, opened 2023-11-25
Handle blank exit codes in probes

bfuzzy1:


makk94:
> @makk94 you're right I reverted nocturnal change.
> 
> Is it worth it to log seperately on it's own return code?

If this cache the handle thing works, we shoudl get legitimate exit codes and we shouldn't need to special case anything. If we don't want to do that, then I have no opinions. I'm not sure what the description of a new exit code would be though ("we couldn't get any data but we think you're fine"? lol), but from the github issue description, I'm not even sure that we can assume empty exit code is protected

bfuzzy1:
> > @makk94 you're right I reverted nocturnal change.
> > Is it worth it to log seperately on it's own return code?
> 
> If this cache the handle thing works, we shoudl get legitimate exit codes and we shouldn't need to special case anything. If we don't want to do that, then I have no opinions. I'm not sure what the description of a new exit code would be though ("we couldn't get any data but we think you're fine"? lol), but from the github issue description, I'm not even sure that we can assume empty exit code is protected

Yea I hadn't thought that one all the way through i.e., new exit code. Did you still have your old cache handle PR?

makk94:
> > > @makk94 you're right I reverted nocturnal change.
> > > Is it worth it to log seperately on it's own return code?
> > 
> > 
> > If this cache the handle thing works, we shoudl get legitimate exit codes and we shouldn't need to special case anything. If we don't want to do that, then I have no opinions. I'm not sure what the description of a new exit code would be though ("we couldn't get any data but we think you're fine"? lol), but from the github issue description, I'm not even sure that we can assume empty exit code is protected
> 
> Yea I hadn't thought that one all the way through i.e., new exit code. Did you still have your old cache handle PR?

Mine was https://github.com/preludeorg/libraries/pull/658/files - tho re-reading the github issue, mine probs won't work since it still uses that `Wait` flag instead of a separate `Wait-Process`

makk94:
> Check my latest commit. @gerbsec put that together.

Makes sense to me. Assume it was tested so we're at least confident it doesn't break anything?

shell/probe/raindrop.ps1
```diff
@@ -4,9 +4,16 @@ function Execute {
     Param([String]$File)
 
     try {
-        $R = (Start-Process -FilePath $File -Wait -NoNewWindow -PassThru).ExitCode
-        $Code = if (Test-Path $File) {$R} Else {127}
-        return $Code
+        $process = Start-Process -FilePath $File -NoNewWindow -PassThru
+        $Handle = $process.Handle
```
makk94, line 8: nit: lowercase the `H`

shell/probe/raindrop.ps1
```diff
@@ -4,9 +4,16 @@ function Execute {
...
+        $process.WaitForExit()
+        $exitCode = $process.ExitCode
+
+        if (-not (Test-Path $File)) {
```
makk94, line 12: nit: prefer positive conditions instead of negative unless there's a good reason (`if Test-Path` rather than `if not Test-Path`)

bfuzzy1:
> > Check my latest commit. @gerbsec put that together.
> 
> Makes sense to me. Assume it was tested so we're at least confident it doesn't break anything?

@gerbsec you said you tested this correct with no issues?

gerbsec:
> > > Check my latest commit. @gerbsec put that together.
> > 
> > 
> > Makes sense to me. Assume it was tested so we're at least confident it doesn't break anything?
> 
> @gerbsec you said you tested this correct with no issues?

Yeah I've tested it. I haven't tested it within the probe, but the code doesn't effect anything else I don't see how the probe would change anything. Other wise its working as intended.

bfuzzy1:
> > > > Check my latest commit. @gerbsec put that together.
> > > 
> > > 
> > > Makes sense to me. Assume it was tested so we're at least confident it doesn't break anything?
> > 
> > 
> > @gerbsec you said you tested this correct with no issues?
> 
> Yeah I've tested it. I haven't tested it within the probe, but the code doesn't effect anything else I don't see how the probe would change anything. Other wise its working as intended.

@gerbsec Oh awesome! Can you test the latest commit on this PR? just want to make sure my changes are 100% good to go.

makk94 (approved):
Assuming it's retested, lgtm!

makk94:
Do we need to version this up?

bfuzzy1:
> Do we need to version this up?



> Assuming it's retested, lgtm!

Would feel way more comfortable if you were able to test it as well. If you don't mind. :) There's no rush on this change and I want to be extra sure as we take over probe code. Also, I don't think we version anything with a probe change? Least I didn't think we did.

makk94:
@bfuzzy1 Ran all the tests on a probe with this code, and all looked normal to me!

PR #918, opened 2023-12-08
fix: add autocomplete for profile

packplusplus:
Brought your pr back again


stephanwampouille:
let me know when you want a review on this

packplusplus:
> let me know when you want a review on this

I've been using it locally with success. To be clear, it's something mahina original wrote, and I just resurrected it. The fix commit was because an expired option snuck in when I merged master back in. You can review it whenever.

stephanwampouille (approved):

PR #919, opened 2023-12-11
Template README On Test Creation

smprelude:
Added functionality to the cli's `build` feature that writes a templated `README.md` markdown file (also included in this update) to the new test's directory when a user creates a new test with `prelude build create-test <opts>`.

Currently, this functionality is broken up for readability, but can be condensed if need be, for example by combining both file writes into one `with open` call. I think it should probably remain separated, though.

Sample cli output screens: 
Screenshot 2023-12-08 210558
Screenshot 2023-12-08 210610

### DIFF
--- a/python/cli/prelude_cli/templates/README.md
+++ b/python/cli/prelude_cli/templates/README.md
@@ -1,0 +1,28 @@
+# $NAME
+
+This section should provide a description of the test, providing some background information (such as historical in-the-wild usage by threat actors or APTs), as well as briefly illustrating the test's intention, objective, and techniques utilized to achieve them. This section should NOT be too technical or overly long; a small paragraph should be sufficient.
+
+## How
+
+> Safety: This section should contain a kind of advisory statement which positively declares that the VST does not actually negatively affect the tested system in any way.
+
+Steps:
+
+1. This section enumerates the individual, atomic steps by which ...
+2. ... the probe attempts to achieve the intended effect of the test ...
+3. ... and thereby perform evaluation of any resident security solutions.
+
+Example Output:
+```bash
+[$ID] Starting test at: $TIME
+[$ID] Host is vulnerable, continuing with technique execution
+[$ID] Completed with code: 101
+[$ID] Ending test at: $TIME
+```
+
+## Resolution
+
+If this test fails:
+
+* Bulleted list of recommendations that the testing organization SHOULD consider ...
+* ... based upon the nature of the test, as well as the point of failure.
--- a/python/cli/prelude_cli/templates/template.go
+++ b/python/cli/prelude_cli/templates/template.go
@@ -2,7 +2,7 @@
 ID: $ID
 NAME: $NAME
 UNIT: $UNIT
-CREATED: $CREATED
+CREATED: $TIME
 */
 package main
 
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -36,7 +36,7 @@ def create_test(controller, name, unit, test, techniques, advisory):
         utc_time = str(datetime.now(timezone.utc))
         template_body = pkg_resources.read_text(templates, template)
         template_body = template_body.replace('$ID', t['id'])
-        template_body = template_body.replace('$NAME', name)
+        template_body = template_body.replace('$NAME', t['name'])
         template_body = template_body.replace('$UNIT', unit or '')
         template_body = template_body.replace('$TIME', utc_time)
         
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -61,6 +61,8 @@ def create_test(controller, name, unit, test, techniques, advisory):
         with open(test_dir, 'w', encoding='utf8') as test_code:
             test_code.write(testTemplate)
 
+        print_json(data=t)
+
         """ Mark up, upload, and write the test README template. """
         readmeTemplate = pkg_resources.read_text(templates, 'README.md')
         readmeTemplate = readmeTemplate.replace('$NAME', name)
@@ -70,7 +72,9 @@ def create_test(controller, name, unit, test, techniques, advisory):
         with Spinner(description='Applying default template to new test'):
             controller.upload(test_id=t['id'], filename='README.md', data=readmeTemplate.encode('utf-8'))
 
-        with open(test_dir, 'w', encoding='utf8') as test_readme:
+        readme_dir = PurePath(t['id'], 'README.md')
+
+        with open(readme_dir, 'w', encoding='utf8') as test_readme:
             test_readme.write(readmeTemplate)
             t['attachments'] = ['README.md']
 
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -43,25 +43,24 @@ def create_test(controller, name, unit, test, techniques, advisory):
 
     def create_template(template, name):
         utc_time = str(datetime.now(timezone.utc))
-        template = pkg_resources.read_text(templates, template)
-        template = template.replace('$ID', t['id'])
-        template = template.replace('$NAME', name)
-        template = template.replace('$UNIT', unit or '')
-        template = template.replace('$TIME', utc_time)
+        template_body = pkg_resources.read_text(templates, template)
+        template_body = template_body.replace('$ID', t['id'])
+        template_body = template_body.replace('$NAME', name)
+        template_body = template_body.replace('$UNIT', unit or '')
+        template_body = template_body.replace('$TIME', utc_time)
         
         with Spinner(description='Applying default template to new test'):
-            controller.upload(test_id=t['id'], filename=name, data=template.encode('utf-8'))
+            controller.upload(test_id=t['id'], filename=name, data=template_body.encode('utf-8'))
             t['attachments'] += [name]
 
         dir = PurePath(t['id'], name)
-        Path(t['id']).mkdir(parents=True, exist_ok=True)
         
         with open(dir, 'w', encoding='utf8') as code:
-            code.write(template)
+            code.write(template_body)
 
     if not test:
-        basename = f'{t["id"]}.go'
-        create_template(template='template.go', name=basename)
+        Path(t['id']).mkdir(parents=True, exist_ok=True)
+        create_template(template='template.go', name=f'{t["id"]}.go')
         create_template(template='README.md', name='README.md')
 
     print_json(data=t)
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -31,6 +31,7 @@ def build(ctx):
 @click.pass_obj
 @handle_api_error
 def create_test(controller, name, unit, test, techniques, advisory):
+    """ Create or update a security test """
     def create_template(template, name):
         utc_time = str(datetime.now(timezone.utc))
         template_body = pkg_resources.read_text(templates, template)
@@ -39,7 +40,6 @@ def create_test(controller, name, unit, test, techniques, advisory):
         template_body = template_body.replace('$UNIT', unit or '')
         template_body = template_body.replace('$TIME', utc_time)
 
-    """ Create or update a security test """
     with Spinner(description='Creating new test'):
         t = controller.create_test(
             name=name,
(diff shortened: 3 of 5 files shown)


python/cli/prelude_cli/views/build.py
```diff
@@ -42,22 +42,41 @@ def create_test(controller, name, unit, test, techniques, advisory):
...
     if not test:
+        """ Mark-up, upload, and write the test file template. """
         basename = f'{t["id"]}.go'
-        template = pkg_resources.read_text(templates, 'template.go')
-        template = template.replace('$ID', t['id'])
-        template = template.replace('$NAME', name)
-        template = template.replace('$UNIT', unit or '')
-        template = template.replace('$CREATED', str(datetime.utcnow()))
+        utcTime = str(datetime.utcnow())
+        testTemplate = pkg_resources.read_text(templates, 'template.go')
```
stephanwampouille, line 48: Can you change the var names to be more consistent with how the codebase handles python vars. Words separated by underscores.

```suggestion
        test_template = pkg_resources.read_text(templates, 'template.go')
```

stephanwampouille (changes requested):
Overall looks good, and I like the idea of populating the template with more context for a user. My main request is just making sure we are being consistent with variable syntaxing in the codebase so once that is adjusted I will re-review!

python/cli/prelude_cli/views/build.py
```diff
@@ -42,22 +42,41 @@ def create_test(controller, name, unit, test, techniques, advisory):
...
 
     if not test:
+        """ Mark-up, upload, and write the test file template. """
         basename = f'{t["id"]}.go'
-        template = pkg_resources.read_text(templates, 'template.go')
-        template = template.replace('$ID', t['id'])
-        template = template.replace('$NAME', name)
-        template = template.replace('$UNIT', unit or '')
-        template = template.replace('$CREATED', str(datetime.utcnow()))
+        utc_time = str(datetime.utcnow())
```
packplusplus, line 47: `utcnow` is on it's way out, we started removing it from the backend code. Since you're touching this, maybe make it timezone aware while you're at it? 

* https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated
* example of a tz aware datetime: https://github.com/preludeorg/backbone/blob/5a595f551202d1e932c52f815954670c1e69ae86/backbone/objects/standard.py#L73C18-L73C18

packplusplus:
Just my two cents, I'll defer to Stephan for approval.

python/cli/prelude_cli/templates/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+This section should describe the test, provide some background information (such as historical in-the-wild usage by threat actors or APTs), and briefly illustrate the test's intention, objective, and techniques utilized to achieve them. This section should NOT be too technical or overly long; a small paragraph should be sufficient.
+
+## How
+
+ > Safety: This section should contain an advisory statement that positively declares that the VST does not negatively affect the tested system in any way.
+
+Steps:
+
+ 1. This section enumerates the individual, atomic steps by which ...
```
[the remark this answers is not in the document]
smprelude: Oh whoops, I guess I didn't save my comment before resolving. My bad!

This is neat, but I don't think we necessarily should presume user familiarity with markdown formatting and/or try to teach them about it. Leaving as-is.

python/cli/prelude_cli/views/build.py
```diff
@@ -5,7 +5,7 @@
 import importlib.resources as pkg_resources
 
 from rich import print_json
-from datetime import datetime
+from datetime import datetime, timezone
```
makk94, line 8: can you bump this down a line (we order our imports by length. we know its weird ha)

python/cli/prelude_cli/views/build.py
```diff
@@ -42,22 +42,41 @@ def create_test(controller, name, unit, test, techniques, advisory):
...
         test_dir = PurePath(t['id'], basename)
         Path(t['id']).mkdir(parents=True, exist_ok=True)
         
         with open(test_dir, 'w', encoding='utf8') as test_code:
-            test_code.write(template)
+            test_code.write(test_template)
+
+        print_json(data=t)
+
+        """ Mark up, upload, and write the test README template. """
```
makk94, line 66: nit: rm the comments (here and line 45) - code is self explanatory

python/cli/prelude_cli/views/build.py
```diff
@@ -42,22 +42,41 @@ def create_test(controller, name, unit, test, techniques, advisory):
...
+        readme_template = pkg_resources.read_text(templates, 'README.md')
+        readme_template = readme_template.replace('$NAME', name)
+        readme_template = readme_template.replace('$ID', t['id'])
+        readme_template = readme_template.replace('$TIME', utc_time)
+
+        with Spinner(description='Applying default template to new test'):
+            controller.upload(test_id=t['id'], filename='README.md', data=readme_template.encode('utf-8'))
+
+        readme_dir = PurePath(t['id'], 'README.md')
```
makk94, line 75: nit: this is the `readme_file` not `readme_dir`, right?

python/cli/prelude_cli/views/build.py
```diff
@@ -42,22 +42,41 @@ def create_test(controller, name, unit, test, techniques, advisory):
...
             t['attachments'] = [basename]
 
         test_dir = PurePath(t['id'], basename)
         Path(t['id']).mkdir(parents=True, exist_ok=True)
         
         with open(test_dir, 'w', encoding='utf8') as test_code:
-            test_code.write(template)
+            test_code.write(test_template)
+
+        print_json(data=t)
```
makk94, line 64: Can we have a single print at the end with both attachments? (rather than two separate prints)

python/cli/prelude_cli/views/build.py
```diff
@@ -42,22 +42,41 @@ def create_test(controller, name, unit, test, techniques, advisory):
...
+
+        """ Mark up, upload, and write the test README template. """
+        readme_template = pkg_resources.read_text(templates, 'README.md')
```
makk94, line 67: This block is so similar to the block above, it'd be nice to have an inner func to dedupe the code so the "main" code can just be

```
create_attachement(template='template.go', name=basename)
create_attachement(template='README.md', name='README.md')
```

python/cli/prelude_cli/views/build.py
smprelude: like insert \n between lines 7 and 8?

makk94:
```
from rich ...
from pathlib import Path, PurePath
from datetime import datetime, timezone
```

It's a little weird, but until we fix all the imports everywhere, i'd like to stay consistent

python/cli/prelude_cli/views/build.py
smprelude: done

python/cli/prelude_cli/views/build.py
smprelude: ah gotcha. done and done :)

python/cli/prelude_cli/views/build.py
smprelude: done. remove this call, use the second one at the bottom.

python/cli/prelude_cli/views/build.py
smprelude: eh, i'm going to punt on this for now. i agree it may be nice to have a helper func to tidy this section up, but don't think it's strictly necessary at this very moment.

makk94: I very much take the other view ha unless there is a _very_ good reason to not have clean code, default to doing it the right way the first time

Our code base right now is super slim, has next to no tech debt, and I'd love to keep it that way as long as possible

python/cli/prelude_cli/templates/template.go
```diff
@@ -2,7 +2,7 @@
 ID: $ID
 NAME: $NAME
 UNIT: $UNIT
-CREATED: $CREATED
+CREATED: $TIME
```
makk94, line 5: It's probably worth updating the template/test in the javascript test code too (just to keep it all consistent)
* https://github.com/preludeorg/libraries/blob/master/javascript/sdk/test/sdk.test.ts#L154
* https://github.com/preludeorg/libraries/blob/master/javascript/sdk/test/templates/template.go#L5

python/cli/prelude_cli/views/build.py
```diff
@@ -41,23 +41,28 @@ def create_test(controller, name, unit, test, techniques, advisory):
             advisory=advisory
         )
 
-    if not test:
-        basename = f'{t["id"]}.go'
-        template = pkg_resources.read_text(templates, 'template.go')
+    def create_template(template, name):
+        utc_time = str(datetime.now(timezone.utc))
+        template = pkg_resources.read_text(templates, template)
```
makk94, line 46: nit: will you call the input param something other than `template` just so we don't have the variable name reuse

python/cli/prelude_cli/views/build.py
```diff
@@ -41,23 +41,28 @@ def create_test(controller, name, unit, test, techniques, advisory):
...
         
         with Spinner(description='Applying default template to new test'):
-            controller.upload(test_id=t['id'], filename=basename, data=template.encode('utf-8'))
-            t['attachments'] = [basename]
+            controller.upload(test_id=t['id'], filename=name, data=template.encode('utf-8'))
+            t['attachments'] += [name]
 
-        test_dir = PurePath(t['id'], basename)
+        dir = PurePath(t['id'], name)
         Path(t['id']).mkdir(parents=True, exist_ok=True)
```
makk94, line 57: no need for this `mkdir` to be in the inner function - i'd pull it out into the main function logic

python/cli/prelude_cli/views/build.py
```diff
@@ -41,23 +41,28 @@ def create_test(controller, name, unit, test, techniques, advisory):
...
         
-        with open(test_dir, 'w', encoding='utf8') as test_code:
-            test_code.write(template)
+        with open(dir, 'w', encoding='utf8') as code:
+            code.write(template)
+
+    if not test:
+        basename = f'{t["id"]}.go'
```
makk94, line 63: nit: you can have one less line if you inline this :)

smprelude: savings :D

python/cli/prelude_cli/views/build.py
smprelude: i switched up the name of the variable, instead.

python/cli/prelude_cli/templates/template.go
smprelude: done!

python/cli/prelude_cli/views/build.py
```diff
@@ -41,23 +41,27 @@ def create_test(controller, name, unit, test, techniques, advisory):
...
         )
 
-    if not test:
-        basename = f'{t["id"]}.go'
-        template = pkg_resources.read_text(templates, 'template.go')
-        template = template.replace('$ID', t['id'])
-        template = template.replace('$NAME', name)
-        template = template.replace('$UNIT', unit or '')
-        template = template.replace('$CREATED', str(datetime.utcnow()))
+    def create_template(template, name):
```
makk94, line 44: move the inner func to the top of the method pls

python/cli/prelude_cli/views/build.py
```diff
@@ -31,6 +31,14 @@ def build(ctx):
 @click.pass_obj
 @handle_api_error
 def create_test(controller, name, unit, test, techniques, advisory):
```
makk94, line 33: well. still keep the comment at the top ha
```
def blah(...):
""" function comment """
    def inner_blah(...):
        # do stuff
  
    # main function stuff
    print(result)
```

python/cli/prelude_cli/views/build.py
```diff
@@ -40,24 +48,20 @@ def create_test(controller, name, unit, test, techniques, advisory):
...
-
-    if not test:
-        basename = f'{t["id"]}.go'
-        template = pkg_resources.read_text(templates, 'template.go')
-        template = template.replace('$ID', t['id'])
-        template = template.replace('$NAME', name)
-        template = template.replace('$UNIT', unit or '')
-        template = template.replace('$CREATED', str(datetime.utcnow()))
         
         with Spinner(description='Applying default template to new test'):
```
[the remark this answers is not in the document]
smprelude: oh whoops. you're right. moved chunk to its rightful home :)

makk94 (approved):

smprelude:
@stephanwampouille poke for merge! test output for latest changes on build attached.
image
image

stephanwampouille (approved):
Looks good, and great attention to detail @makk94 ! Nice work @smprelude !

PR #924, opened 2023-12-19
det-949: add update-user to CLI and subscribable events 

makk94:
PR bundle:
* https://github.com/preludeorg/service/pull/1568
* https://github.com/preludeorg/backbone/pull/701
* https://github.com/preludeorg/libraries/pull/924

### DIFF
--- a/python/sdk/prelude_sdk/models/codes.py
+++ b/python/sdk/prelude_sdk/models/codes.py
@@ -182,7 +182,9 @@ class AuditEvent(Enum, metaclass=MissingItem):
     UPDATE_ACCOUNT = 13
     UPDATE_ENDPOINT = 14
     UPDATE_TEST = 15
+    UPDATE_USER = 17
     UPLOAD_TEST_ATTACHMENT = 16
+    # Next value: 18
 
     @classmethod
     def _missing_(cls, value):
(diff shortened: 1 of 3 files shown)


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -182,7 +182,9 @@ class AuditEvent(Enum, metaclass=MissingItem):
     UPDATE_ACCOUNT = 13
     UPDATE_ENDPOINT = 14
     UPDATE_TEST = 15
+    UPDATE_USER = 17
```
stephanwampouille, line 185: Shouldn't this come after 16?

stephanwampouille (changes requested):
What are the audit events used for :)

python/sdk/prelude_sdk/models/codes.py
makk94: People (maybe users, mostly devs (ie. me)) scan lists easier if they're in alphabetical order. Since the numbers mean nothing here, I'm prioritizing ordering the names instead of the numbers

makk94:
> What are the audit events used for :)

You can subscribe to an audit event and get real time emails when an event you care about happens. Ex. you can subscribe to CREATE_TEST events and you'll get an email every time a new test is created in your account

stephanwampouille (approved):
looks good!

PR #927, opened 2023-12-27
det-950: add slug to account

makk94:
PR bundle:
- https://github.com/preludeorg/backbone/pull/705
- https://github.com/preludeorg/service/pull/1579
- https://github.com/preludeorg/libraries/pull/927

### DIFF
--- a/python/cli/prelude_cli/views/iam.py
+++ b/python/cli/prelude_cli/views/iam.py
@@ -24,7 +24,7 @@ def register_account(controller):
     email = click.prompt('Enter your email')
     name = click.prompt('(Optional) Enter your name', default='', show_default=False)
     company = click.prompt('(Optional) Enter your associated company', default='', show_default=False)
-    slug = click.prompt('(Optional) Enter an account slug', default='', show_default=False)
+    slug = click.prompt('(Optional) Enter an unique human-readable identifier for your account', default='', show_default=False)
     with Spinner(description='Creating new account'):
         data = controller.new_account(user_email=email, user_name=name, company=company, slug=slug)
     print_json(data=data)
@@ -37,7 +37,7 @@ def register_account(controller):
               default=None, show_default=False,
               type=click.Choice([m.name for m in Mode], case_sensitive=False))
 @click.option('-c', '--company', help='provide your associated company', default=None, show_default=False, type=str)
-@click.option('-s', '--slug', help='provide your unique slug', default=None, show_default=False, type=str)
+@click.option('-s', '--slug', help='provide a unique human-readable identifier for your account', default=None, show_default=False, type=str)
 @click.pass_obj
 @handle_api_error
 def update_account(controller, mode, company, slug):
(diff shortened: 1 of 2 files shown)


python/cli/prelude_cli/views/iam.py
```diff
@@ -24,8 +24,9 @@ def register_account(controller):
     email = click.prompt('Enter your email')
     name = click.prompt('(Optional) Enter your name', default='', show_default=False)
     company = click.prompt('(Optional) Enter your associated company', default='', show_default=False)
+    slug = click.prompt('(Optional) Enter an account slug', default='', show_default=False)
```
packplusplus, line 27: '' here but none in click option.

python/sdk/prelude_sdk/controllers/iam_controller.py
```diff
@@ -47,13 +49,15 @@ def purge_account(self):
...
     @verify_credentials
-    def update_account(self, mode: Mode = None, company: str = None):
+    def update_account(self, mode: Mode = None, company: str = None, slug: str = None):
         """ Update properties on an account """
         body = dict()
         if mode is not None:
             body['mode'] = mode.name
         if company is not None:
             body['company'] = company
+        if slug is not None:
```
packplusplus, line 59: why isn't this just "if slug" ? To match the previous style?

python/cli/prelude_cli/views/iam.py
makk94: `click.prompt` needs an input (it'll just keep asking the same question if I default None). Ultimately, empty string and None will be treated the same way in sdk (`if slug: <blah>`)

python/sdk/prelude_sdk/controllers/iam_controller.py
makk94: There are two things that a user could want to do:
1) update other fields and leave slug as is
2) set slug to None

In the CLI:
The first is done by not using the `--slug` flag at all, which defaults the `slug` value to None.
To differentiate from that, the second is going to be done by setting the `--slug` flag to `''` 

So in the SDK:
If `slug` is None, we don't set the `slug` field at all in our API request, which means don't touch that field in the BE
If `slug` is empty string, the `slug` field will be in our API request, and the empty string will convert to `None` in the BE

It isn't the cleanest solution, but it works. If you can think of a better way to differentiate, I'm open!

stephanwampouille:
Looks good but I wonder if `slug` is too generic of a word in this case that without more docs a user won't understand its use case? An argument can be made that this step will be a walkthrough with Pete and Joe but want to see your opinions.

makk94:
> Looks good but I wonder if `slug` is too generic of a word in this case that without more docs a user won't understand its use case? An argument can be made that this step will be a walkthrough with Pete and Joe but want to see your opinions.

I updated the help sentences to say "human-readable unique identifier for your account" instead of "slug". Think that's good?

stephanwampouille (approved):
I think that's definitely better!

PR #934, opened 2024-01-08
det-975: get probe sleep interval from BE

makk94:
Also, fix upgrade path for raindrop :( 

If when this goes in, I'll update probe_tasks probe version as well to trigger upgrades (for nocturnal anyway)

PR bundle:
- https://github.com/preludeorg/backbone/pull/719
- https://github.com/preludeorg/service/pull/1601
- https://github.com/preludeorg/libraries/pull/934

### DIFF
--- a/shell/probe/nocturnal.sh
+++ b/shell/probe/nocturnal.sh
@@ -24,10 +24,6 @@ do
         sleep ${sleep_sec:-1800}
 
         # Option 2. try/catch method. I couldn't find a way to do sneaky bash injection, and this might is probably slightly more efficient
-        sleep $task 2>/dev/null
-        if [ $? = 1 ];then
-          echo "Invalid sleep time: $task"
-          sleep 1800
-        fi
+        sleep $task 2>/dev/null || (echo "Invalid sleep time: $task" && sleep 1800)
     fi
 done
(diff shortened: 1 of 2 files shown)


packplusplus (commented):
I don't love it. 

In the case of stop or upgrade, the agent was saying "did i get the value stop, if so exit"

In this case we're just grabbing the value, not sanity checking it, and using it. 

So the nocturnal one could 100% just be like "5; sudo reboot" and I think it would execute that code. Not sure about the powershell one, but similar concern there.

makk94:
@packplusplus you have not been requested to review this PR

packplusplus:
*cough* draft *cough*

shell/probe/nocturnal.sh
```diff
@@ -18,6 +18,12 @@ do
     else
         rm -rf $vst
         unset dat
-        sleep 3600
+
+        # Option 1. Explicitly check that sleep is an int. Otherwise default to 1800s
```
bfuzzy1, line 22: I am a fan of the explicity you have in option 1. I would approve that option.

makk94: kewl. patrick agrees

out of curiosity, can you think of a way to take advantage of option 2?

bfuzzy1: I honestly can not think of anything worth the effort.

packplusplus (approved):
I'm convinced. Thank you for sanitizing your inputs.

PR #938, opened 2024-01-12
fix: [DET-1022] Fix probe deploy notifcations.

ohlol:
The notifcations were using `GITHUB_REF` which is not valid (cf. https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables).

They should use `github.ref`.


packplusplus (approved):
Hooray! Thank you.

PR #945, opened 2024-01-30
Slightly more robust xor functionality

smprelude:
In this PR:
Modified the `XorEncrypt` and `XorDecrypt` functions in Endpoint.go to utilize different encoding logic over a longer key, more closely resembling something observable ITW. 

This produces data output which is slightly less trivial for endpoint security appliances/solutions to decode, making the tests which employ this functionality that much more difficult and therefore useful & impactful.


smprelude:
Forgot to mention that this change does not require any changes made to the one or two tests which currently make use of the xor functions.

PR #955, opened 2024-02-20
det-1107: removing advisories

makk94:
* remove `advisory` field from request and response in CreateTest / UpdateTest
* remove ListAdvisories API
* remove `advisories` activity view

### DIFF
--- a/python/sdk/tests/requirements.txt
+++ b/python/sdk/tests/requirements.txt
@@ -1,3 +1,4 @@
 pytest
 pytest-order
+python-dateutil
 ./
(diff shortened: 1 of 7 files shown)


python/sdk/tests/requirements.txt
```diff
@@ -1,3 +1,4 @@
 pytest
 pytest-order
+python-dateutil
```
stephanwampouille, line 3: yay

stephanwampouille (approved):
looks good, ill make a JS one in a bit!

PR #961, opened 2024-02-29
det-1130: add purge option to delete-test and delete-threat

keitharobertson:


python/sdk/prelude_sdk/controllers/build_controller.py
```diff
@@ -49,10 +49,11 @@ def update_test(self, test_id, name=None, unit=None, technique=None):
         raise Exception(res.text)
 
     @verify_credentials
-    def delete_test(self, test_id):
+    def delete_test(self, test_id, purge):
         """ Delete an existing test """
+        purge_arg = '?purge=true' if purge else ''
```
makk94, line 54: can you use the `params` param instead of building your own query

makk94 (approved):
looks good! assuming the sdk test changes are coming in a separate PR?

stephanwampouille:
I know I am late, but is purge used to tombstone a test/threat?

keitharobertson:
> I know I am late, but is purge used to tombstone a test/threat?

purge is used to completely delete a test/threat. Tombstone marks it as deleted but doesn't actually delete anything. If you set purge=true, it will delete the row out of the database and delete the files off S3.

PR #962, opened 2024-03-01
det-1130: add test for purging a test

keitharobertson:
* test purging a test
* fix existing test tombstone test
* fix expected probe sleep time in iam test

I didn't add a test for deleting/purging a threat because everything in the SDK threats appears to be done as a regular user test and threats can only be created / deleted by the prelude account. Looks like there are no create threat tests in there already and I assumed that is why. If that assumption is incorrect let me know and I'll see about adding some tests around threats.

### DIFF
--- a/python/sdk/tests/test_build.py
+++ b/python/sdk/tests/test_build.py
@@ -108,13 +108,20 @@ class TestBuild:
 
     @pytest.mark.order(-2)
     def test_delete_test(self, unwrap):
-        unwrap(self.build.delete_test)(self.build, test_id=pytest.test_id)
+        unwrap(self.build.delete_test)(self.build, test_id=pytest.test_id, purge=False)
         res = unwrap(self.detect.get_test)(self.detect, test_id=pytest.test_id)
         pytest.expected_test['tombstoned'] = res['tombstoned']
 
         diffs = check_dict_items(pytest.expected_test, res)
         assert not diffs, json.dumps(diffs, indent=2)
         assert parse(res['tombstoned']) <= datetime.utcnow() + timedelta(minutes=1)
+
+    def test_purge_test(self, unwrap):
+        purge_test_id = str(uuid.uuid4())
+        unwrap(self.build.create_test)(self.build, test_id=purge_test_id, name=self.name, unit=self.unit)
+        unwrap(self.build.delete_test)(self.build, test_id=purge_test_id, purge=True)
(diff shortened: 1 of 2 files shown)


python/sdk/tests/test_build.py
```diff
@@ -108,14 +108,21 @@ def test_download(self, unwrap):
...
-        unwrap(self.build.delete_test)(self.build, test_id=pytest.test_id)
+        unwrap(self.build.delete_test)(self.build, test_id=pytest.test_id, purge=False)
         res = unwrap(self.detect.get_test)(self.detect, test_id=pytest.test_id)
         pytest.expected_test['tombstoned'] = res['tombstoned']
 
         diffs = check_dict_items(pytest.expected_test, res)
         assert not diffs, json.dumps(diffs, indent=2)
         assert parse(res['tombstoned']) <= datetime.utcnow() + timedelta(minutes=1)
 
+    def test_purge_test(self, unwrap):
```
makk94, line 119: i think you probably need to mark this one as `@pytest.mark.order(-2)` and the one above as `@pytest.mark.order(-3)` to make them run correctly

makk94, line 119: this is weird to throw in one random semi-independent test amongst everything. prefer you just tack it on to the rest of the test suite (purge the just-tombstoned test above), but if you're keen for independent tests, start some new files that we can slowly start porting to

keitharobertson: 👍 I was concerned deleting it would break other tests that depend on the test existing

makk94: the `@pytest.mark.order(-2)` on `test_delete_test` means "run it second to last" (purge_account is last). so you can adjust that one and put the order decorator on this one so that purge_test runs second to last and delete_test runs third to last

keitharobertson: ah ok that should work then

makk94:
> I didn't add a test for deleting/purging a threat because everything in the SDK threats appears to be done as a regular user test and threats can only be created / deleted by the prelude account. Looks like there are no create threat tests in there already and I assumed that is why. If that assumption is incorrect let me know and I'll see about adding some tests around threats.

yeah, your read is correct. when we open up create/update/deleteThreat to the public, then we'll add those tests here

makk94 (approved):

PR #965, opened 2024-03-05
det-1151: update schedule/unschedule python SDK to take a list of tests/threats

makk94:
CLI staying with single test/threat for now. lists of dicts are awkward in CLIs


stephanwampouille (approved):
love the example blurb!

PR #967, opened 2024-03-07
fix python sdk tests

makk94:
* alphabetize imports
* the threat I'm using in GetThreat had metadata updates
* schedule params were wrong. so was the response I was expecting

I ran them locally this time and all are passing

### DIFF
--- a/python/sdk/tests/test_detect.py
+++ b/python/sdk/tests/test_detect.py
@@ -1,14 +1,14 @@
 import json
 import pytest
 import requests
+from datetime import datetime, timedelta
 
 from dateutil.parser import parse
-from datetime import datetime, timedelta
-from prelude_sdk.models.codes import RunCode
-from prelude_sdk.controllers.iam_controller import IAMController
-from prelude_sdk.controllers.detect_controller import DetectController
-
 from testutils import *
+
+from prelude_sdk.controllers.detect_controller import DetectController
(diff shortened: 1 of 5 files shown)


python/sdk/tests/test_detect.py
```diff
@@ -62,8 +62,10 @@ def test_update_endpoint(self, unwrap):
         assert not diffs, json.dumps(diffs, indent=2)
 
     def test_schedule_threat(self, unwrap):
-        res = unwrap(self.detect.schedule)(self.detect, [dict(threat_id=pytest.threat_id, run_code=RunCode.DAILY)])
-        assert res['id'] == pytest.threat_id
+        res = unwrap(self.detect.schedule)(self.detect, [dict(threat_id=pytest.threat_id, code=RunCode.DAILY.name)])
+        assert 1 == len(res), json.dumps(res, indent=2)
+        diffs = check_dict_items(dict(threat=pytest.threat_id, run_code=RunCode.DAILY.value, tag=None), res[0])
```
keitharobertson, line 67: so the request uses `code` but the response has `run_code`?

makk94: yeah, that's not great is it...

keitharobertson: if not too late I think we should fix that

keitharobertson (approved):

PR #970, opened 2024-03-15
det-1225: CLI for IOAs

makk94:


stephanwampouille:
Looks good! One question, for create_detection it looks like you cannot create a detection if both `rules` and `rules_file` is set it will fail, however you can create a detection with a `rules_file` and upload `rules` after?

makk94:
> Looks good! One question, for create_detection it looks like you cannot create a detection if both `rules` and `rules_file` is set it will fail, however you can create a detection with a `rules_file` and upload `rules` after?

Those two flags are two ways of providing the same information.
You could do:
```
prelude build create-detection CROWDSTRIKE -t c682461c-b5c7-4e21-be34-8aecf692cb2f  -r '{"name":"mahina", "ruletype_id":"1", "field_values": [
            {"label":"Grandparent Image Filename",
            "name": "GrandparentImageFilename",
            "type": "excludable",
            "values": [
                {
                    "label": "include",
                    "value": "mahina.exe"
                }
            ]
        }]}'
```

or you could make a file called `rule.json` with the contents:
```
{"name":"mahina", "ruletype_id":"1", "field_values": [
            {"label":"Grandparent Image Filename",
            "name": "GrandparentImageFilename",
            "type": "excludable",
            "values": [
                {
                    "label": "include",
                    "value": "mahina.exe"
                }
            ]
        }]}
```
and then create the detection via:
```
prelude build create-detection CROWDSTRIKE -t c682461c-b5c7-4e21-be34-8aecf692cb2f  --rules_file rule.json
```

python/cli/prelude_cli/views/build.py
```diff
@@ -188,3 +190,59 @@ def delete_threat(controller, threat, purge):
...
+            partner=Control[partner],
+            detection_id=detection_id,
+            rule_id=rule_id
+        )
+    print_json(data=t)
+
+
+@build.command('update-detection')
+@click.argument('detection')
+@click.option('-r', '--rules', help='CRWD detection rules, as string-formatted json', default=None, type=str)
```
keitharobertson, line 222: should update also accept a rules_file to be consistent with create?

keitharobertson (approved):

python/cli/prelude_cli/views/build.py
makk94: yes. done!

keitharobertson (approved):

PR #976, opened 2024-03-28
Add generate threat intel

keitharobertson:
* Generate threat uploads a PDF, polls the status, and writes the output disk
* Add a directory option to create-threat to take threat intel output and create tests and a threat

### DIFF
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -1,5 +1,6 @@
 import importlib.resources as pkg_resources
 import json
+import os
 import re
 import uuid
 from datetime import datetime, timezone
@@ -146,17 +147,21 @@ def upload_attachment(controller, path, test):
 def create_threat(controller, name, published, id, source_id, source, tests, directory):
     """ Create a security threat """
     with Spinner(description='Creating new threat'):
-        if directory:
-            t = controller.create_threat_from_directory(
-                directory=directory,
-                name=name,
-                threat_id=id,
-                source_id=source_id,
-                source=source,
-                published=published,
-            )
-        else:
-            t = controller.create_threat(
+        try:
+            created_tests = list()
+            uploads = list()
+            threat = None
+            if directory:
+                for technique_dir in os.listdir(directory):
+                    with open(f'{directory}/{technique_dir}/test.go', 'r') as f:
+                        go_code = f.read()
+                    with open(f'{directory}/{technique_dir}/config.json', 'r') as f:
+                        config = json.load(f)
+                    test = controller.create_test(name=config['name'], unit=config['unit'], technique=config['technique'])
+                    created_tests.append(test)
+                    uploads.append(controller.upload(test_id=test['id'], filename='test.go', data=go_code.encode()))
+                tests = ','.join([t['id'] for t in created_tests])
+            threat = controller.create_threat(
                 name=name,
                 threat_id=id,
                 source_id=source_id,
@@ -164,7 +169,10 @@ def create_threat(controller, name, published, id, source_id, source, tests, dir
                 published=published,
                 tests=tests,
             )
-    print_json(data=t)
+        except FileNotFoundError as e:
+            raise Exception(e)
+        finally:
+            print_json(data=dict(threat=threat, created_tests=created_tests, uploads=uploads))
 
 
 @build.command('update-threat')
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -140,19 +140,30 @@ def upload_attachment(controller, path, test):
 @click.option('-s', '--source', help='source of threat (ex. www.cisa.gov)', default=None, type=str)
 @click.option('-i', '--source_id', help='ID of the threat, per the source (ex. aa23-075a)', default=None, type=str)
 @click.option('-t', '--tests', help='comma-separated list of test IDs', default=None, type=str)
+@click.option('-d', '--directory', help='directory containing tests and IOAs', default=None, type=click.Path(exists=True, dir_okay=True, file_okay=False))
 @click.pass_obj
 @handle_api_error
-def create_threat(controller, name, published, id, source_id, source, tests):
+def create_threat(controller, name, published, id, source_id, source, tests, directory):
     """ Create a security threat """
     with Spinner(description='Creating new threat'):
-        t = controller.create_threat(
-            name=name,
-            threat_id=id,
-            source_id=source_id,
-            source=source,
-            published=published,
-            tests=tests
-        )
+        if directory:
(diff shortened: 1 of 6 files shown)


python/sdk/prelude_sdk/controllers/build_controller.py
```diff
@@ -102,6 +105,24 @@ def create_threat(self, name, published, threat_id=None, source_id=None, source=
...
+            for technique_dir in os.listdir(directory):
+                with open(f'{directory}/{technique_dir}/test.go', 'r') as f:
+                    go_code = f.read()
+                with open(f'{directory}/{technique_dir}/config.json', 'r') as f:
+                    config = json.load(f)
+                test = self.create_test(name=config['name'], unit=config['unit'], technique=config['technique'])
+                tests.append(test)
+                uploads.append(self.upload(test_id=test['id'], filename='test.go', data=go_code.encode()))
+        except FileNotFoundError as e:
+            raise Exception(e)
```
keitharobertson, line 122: this is weird, but I learned without this FileNotFoundError is assumed to mean the prelude keychain is missing so the output error is real weird if a file is missing here without this.

python/cli/prelude_cli/views/generate.py
```diff
@@ -0,0 +1,54 @@
...
+
+@click.group()
+@click.pass_context
+def generate(ctx):
+    """ Generate tests """
+    ctx.obj = GenerateController(account=ctx.obj)
+
+
+@generate.command('threat-intel')
+@click.argument('file', type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True))
```
makk94, line 19: maybe "threat" or "pdf" ?

python/sdk/prelude_sdk/controllers/build_controller.py
```diff
@@ -102,6 +105,24 @@ def create_threat(self, name, published, threat_id=None, source_id=None, source=
             return res.json()
         raise Exception(res.text)
 
+    @verify_credentials
+    def create_threat_from_directory(self, directory: str, **kwargs):
```
makk94, line 109: light/light-medium opinion: this belongs in the cli (or a cli utils file) more than the sdk since it relies on a particular directory structure that our threat_intel generates

python/sdk/prelude_sdk/controllers/build_controller.py
```diff
@@ -102,6 +105,24 @@ def create_threat(self, name, published, threat_id=None, source_id=None, source=
...
+        tests = list()
+        uploads = list()
+        try:
+            for technique_dir in os.listdir(directory):
+                with open(f'{directory}/{technique_dir}/test.go', 'r') as f:
```
makk94, line 114: why not creating the IOAs too? just try/ignore 403s

keitharobertson: the output of threat intel gen is a yaml. Could probably convert it to whatever we need, but it sounded like we were just ignoring these until we figured out the sigma stuff and then we would have the threat intel stuff output a sigma definition instead.

makk94: ah. fair. kk. i'll update this later then when we figure out the IOA format

python/sdk/prelude_sdk/controllers/build_controller.py
makk94: low key hate that we tied our sdk to the prelude keychain. it's weird

python/sdk/prelude_sdk/controllers/build_controller.py
```diff
@@ -102,6 +105,24 @@ def create_threat(self, name, published, threat_id=None, source_id=None, source=
...
+                    go_code = f.read()
+                with open(f'{directory}/{technique_dir}/config.json', 'r') as f:
+                    config = json.load(f)
+                test = self.create_test(name=config['name'], unit=config['unit'], technique=config['technique'])
+                tests.append(test)
+                uploads.append(self.upload(test_id=test['id'], filename='test.go', data=go_code.encode()))
+        except FileNotFoundError as e:
+            raise Exception(e)
+        threat = self.create_threat(**kwargs, tests=','.join([t['id'] for t in tests]))
+        return {'threat': threat, 'tests': tests, 'uploads': uploads}
```
makk94, line 124: i wonder if we should throw this whole thing in a try/catch so in event of an error half way through, you can still output the uploads that _did_ happen

python/cli/prelude_cli/views/build.py
```diff
@@ -140,20 +141,38 @@ def upload(p: Path):
 @click.option('-s', '--source', help='source of threat (ex. www.cisa.gov)', default=None, type=str)
 @click.option('-i', '--source_id', help='ID of the threat, per the source (ex. aa23-075a)', default=None, type=str)
 @click.option('-t', '--tests', help='comma-separated list of test IDs', default=None, type=str)
+@click.option('-d', '--directory', help='directory containing tests and IOAs', default=None, type=click.Path(exists=True, dir_okay=True, file_okay=False))
```
makk94, line 144: help="directory containing tests and IOAs generated from threat_intel"

python/cli/prelude_cli/views/build.py
```diff
@@ -140,20 +141,38 @@ def upload(p: Path):
...
+            threat = None
+            if directory:
+                for technique_dir in os.listdir(directory):
+                    with open(f'{directory}/{technique_dir}/test.go', 'r') as f:
+                        go_code = f.read()
+                    with open(f'{directory}/{technique_dir}/config.json', 'r') as f:
+                        config = json.load(f)
+                    test = controller.create_test(name=config['name'], unit=config['unit'], technique=config['technique'])
+                    created_tests.append(test)
+                    uploads.append(controller.upload(test_id=test['id'], filename='test.go', data=go_code.encode()))
```
makk94, line 162: nit: make the filename `<test_id>.go` (just so it matches what Prelude does internally and also what the CLI does if you do create_test / upload_attachment flow)

python/cli/prelude_cli/views/generate.py
```diff
@@ -0,0 +1,54 @@
...
+    ctx.obj = GenerateController(account=ctx.obj)
+
+
+@generate.command('threat-intel')
+@click.argument('threat_pdf', type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True))
+@click.argument('output_dir', type=click.Path(dir_okay=True, file_okay=False, writable=True))
+@click.pass_obj
+@handle_api_error
+def generate_threat_intel(controller: GenerateController, threat_pdf: str, output_dir: str):
+    with Spinner('Uploading') as spinner:
```
makk94, line 24: spinner work here is verra nice 👍

makk94 (approved):

PR #977, opened 2024-03-29
Update expected account object to fix sdk test

keitharobertson:


stephanwampouille:
should you add the detections property as well or does it not matter for your tests (I haven't checked the test) ?

keitharobertson:
> should you add the detections property as well or does it not matter for your tests (I haven't checked the test) ?

good point! Now I'm not sure why this passes

keitharobertson:
> > should you add the detections property as well or does it not matter for your tests (I haven't checked the test) ?
> 
> good point! Now I'm not sure why this passes

heh it is looping over the expected items and making sure they are in the returned. So items returned, but not in the expected do not error.

makk94:
> > > should you add the detections property as well or does it not matter for your tests (I haven't checked the test) ?
> > 
> > 
> > good point! Now I'm not sure why this passes
> 
> heh it is looping over the expected items and making sure they are in the returned. So items returned, but not in the expected do not error.

yeah, it's more of a _contains_ test than an exact match. not terribly sophisticated, but i didn't want to deal with difficult fields like timestamps and random ids or whatnot. you like it? ha

makk94:
i have these changes in my PR too :) (just hadn't sent it to you yet)

makk94 (approved):

keitharobertson:
> > > > should you add the detections property as well or does it not matter for your tests (I haven't checked the test) ?
> > > 
> > > 
> > > good point! Now I'm not sure why this passes
> > 
> > 
> > heh it is looping over the expected items and making sure they are in the returned. So items returned, but not in the expected do not error.
> 
> yeah, it's more of a _contains_ test than an exact match. not terribly sophisticated, but i didn't want to deal with difficult fields like timestamps and random ids or whatnot. you like it? ha

Yeah that makes sense!

PR #981, opened 2024-04-05
chore: retry and backoff logic for 429s

packplusplus:
I run into rate limiting when doing big syncs. The sdk will basically throw an http exception and it's up to my client code to make decisions. The exceptions do not have enough information to make a lot of decisions (`Exception(res.text)`). 

I had a couple options here, refactor all the exceptions (sounds like a lot) some kind of decorator that would handle the retries and apply to each function, or moving the requests session to be shared across the class instantiation and use the native retries in requests. I went with the later. No hate on a decorator, I could refactor towards something like that, but the retry backoff logic is already in requests, and we already use requests. 

Thoughts? 

A decorator with some exception handling could make things more clear. The retry logging is all thru the normal logger, and while I played with some magic around tracking class instantiations counts to output when I thought there was a ratelimit, it wasn't good. Or maybe bringing the real exception to the client is the "Right" thing.


makk94:
if this brings you joy, i'm okay with it 👍 
i'd maybe put the default `total` lower just to not hammer our server if client doesn't care

packplusplus:
> if this brings you joy, i'm okay with it 👍 i'd maybe put the default `total` lower just to not hammer our server if client doesn't care

please sir, can i have an approval...

makk94 (approved):

PR #982, opened 2024-04-08
Add unreported exit code

keitharobertson:


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -63,6 +63,7 @@ class ExitCode(Enum):
     MISSING = -1
     UNKNOWN_ERROR = 1
     MALFORMED_TEST = 2
+    UNREPORTED = 3
```
makk94, line 66: and add to ERROR state down around line 130

makk94 (approved):

makk94:
make sure FE is aware of this so they can update their sdk too

makk94:
also update our docs https://docs.preludesecurity.com/docs/understanding-results

stephanwampouille:
will this be going in before the threats deployment?

keitharobertson:
> will this be going in before the threats deployment?

That's the plan unless that causes a problem for you guys

stephanwampouille:
nope that's fine!

PR #985, opened 2024-04-09
Add GetOS helper function to Endpoint

matterpreter:
This MR adds a new helper function that checks to make sure that the test runs on a supported operating system. Currently, we're only supporting Windows, macOS, and Linux.


smprelude (approved):
lgtm

PR #996, opened 2024-04-16
fix: update sdk tests to iso formatted timestamps

makk94:


python/sdk/tests/test_iam.py
```diff
@@ -95,11 +95,11 @@ def test_create_user(self, unwrap):
             dict(
                 handle=self.second_user,
                 permission=Permission.SERVICE.value,
-                # name='Rob',
+                name='Rob',
```
keitharobertson, line 98: just curious why this was commented out before

keitharobertson (approved):

python/sdk/tests/test_iam.py
makk94: i'm sure it got through on accident. sometimes i get paranoid and think that a PASS means that I'm not actually comparing the right objects and i comment things out to force a fail lol

PR #997, opened 2024-04-17
fix: python sdk tests

makk94:


python/sdk/tests/test_partner.py
```diff
@@ -170,7 +171,7 @@ def test_block(self, unwrap, host, edr_id, control, os, platform, policy, policy
             assert f'{pytest.expected_detection["rule"]["title"]} ({pytest.detection_id[:8]}) (0)' == res[0]['rules'][0]['name']
             assert res[0]['rules'][0]['status'] in ['ALREADY_EXISTS', 'CREATED']
         else:
-            assert len(res) == 5
+            assert 5 == len(res)
```
keitharobertson, line 174: do we try to follow `assert expected == actual` in general?

keitharobertson (approved):

python/sdk/tests/test_partner.py
makk94: yeah, it makes life a smidge easier when I'm reading any diffs to know the thing on the right is the actual value

PR #999, opened 2024-04-23
Remove not protected statuses from protected list

keitharobertson:


keitharobertson:
@stephanwampouille @kenrick fyi. `Not relevant` statuses are neither protected or unprotected.

stephanwampouille:
got it!

PR #1003, opened 2024-04-24
Use context to manage timeouts

keitharobertson:
* pass a context with a timeout to start
* pass the context to exec.CommandContext so the child process is killed on timeout

### DIFF
--- a/go/tests/endpoint/endpoint.go
+++ b/go/tests/endpoint/endpoint.go
@@ -3,6 +3,7 @@ package Endpoint
 import (
 	"archive/zip"
 	"bytes"
+	"context"
 	"crypto/aes"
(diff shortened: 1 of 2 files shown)


keitharobertson:
> @keitharobertson @smprelude Where are we with this? Is this a down the road change? This change is very agreeable just needs testing in my opinion.

I had forgotten about this, but agree this is still a change we should make. I'm going to move it out of draft.

keitharobertson:
This requires more discussion on versioning endpoint.go

smprelude:
i'm going to close this PR. it's a decent idea but we've diverged from the impetus which spawned this change in a way that it wouldn't immediately address any longer.

PR #1014, opened 2024-05-14
DET-1084: log one line per test execution in Windows probe

keitharobertson:


shell/probe/raindrop.ps1
```diff
@@ -1,10 +1,10 @@
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 
-function Execute { 
-    Param([String]$File)
+function Execute {
+    Param([String]$File, [String]$StdoutFile, [String]$StderrFile)
 
     try {
-        $R = (Start-Process -FilePath $File -Wait -NoNewWindow -PassThru).ExitCode
+        $R = (Start-Process -FilePath $File -Wait -NoNewWindow -PassThru -RedirectStandardOutput $StdoutFile -RedirectStandardError $StderrFile).ExitCode
```
packplusplus, line 7: Since we're opening this up, maybe settle on variable naming conventions? the rest seem to be lowercase words, not uppercase single letters.

shell/probe/raindrop.ps1
```diff
@@ -33,13 +33,24 @@ while ($true) {
             "dat" = $dat
             "version" = "2.1"
```
packplusplus, line 34: This needs to get bumped.

shell/probe/raindrop.ps1
```diff
@@ -33,13 +33,24 @@ while ($true) {
...
         } -UseBasicParsing -MaximumRedirection 0 -ErrorAction SilentlyContinue
-        
+
         $uuid = $task.content -replace ".*?([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}).*", '$1'
         $auth = $task.content -replace '^[^/]*//([^/]*)/.*', '$1'
 
         if ($uuid -and $auth -eq $ca) {
             Invoke-WebRequest -Uri $task.content -OutFile (New-Item -path "$dir\$uuid.exe" -Force ) -UseBasicParsing
-            $code = Execute "$dir\$uuid.exe"
+            $stdoutTempFile = New-TemporaryFile
```
packplusplus, line 42: Feels like `New-TemporaryFile`s should be in $dir too.

shell/probe/raindrop.ps1
packplusplus, line 7: We should see if we can redirect to the same file, so we can keep ordering? but maybe windows doesn't let you have the same file open for writing in two places.

shell/probe/raindrop.ps1
```diff
@@ -33,13 +33,24 @@ while ($true) {
...
+            $code = Execute "$dir\$uuid.exe" $stdoutTempFile $stderrTempFile
+            $stdout = Get-Content -Path $stdoutTempFile
+            $stdout = if ($stdout) { [string]::Join("; ", $stdout) } else { "" }
+            $stderr = Get-Content -Path $stderrTempFile
+            $stderr = if ($stderr) { [string]::Join("; ", $stderr) } else { "" }
+            if ($stdout -or $stderr) {
+                Write-Output "${stdout};${stderr}"
+            }
+            Remove-Item -Path $stdoutTempFile
+            Remove-Item -Path $stderrTempFile
```
packplusplus, line 53: Do we need to catch these deletes so we report even if we can't delete the file?

shell/probe/raindrop.ps1
makk94: are we confident that no matter what the user's set up, our probe will have write/read access to the temp folder? why not write these files in $dir where we know we have access?

shell/probe/raindrop.ps1
makk94: also https://github.com/preludeorg/service/blob/main/functions/probe_tasks/app.py#L11

shell/probe/raindrop.ps1
makk94: why do we need to write to a temp file? what's the difference between that and just reading stdout/err from the process obj we get back? https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standarderror?view=net-8.0#system-diagnostics-process-standarderror

keitharobertson: for whatever reason `Start-Process` decided to make it very difficult to do so. There are proposed ways to do it, but they risk deadlock under certain conditions:
* https://stackoverflow.com/questions/8761888/capturing-standard-out-and-error-with-start-process
* https://www.reddit.com/r/PowerShell/comments/neqw5a/with_the_startprocess_command_is_there_any_way_to/

redirecting to the file is more foolproof. Agreed on writing to $dir.

shell/probe/raindrop.ps1
keitharobertson: function params are uppercase and everything else is lowercase. No idea why but that's what was in here and I googled it and that does seem to be the powershell convention.

shell/probe/raindrop.ps1
```diff
@@ -33,13 +33,24 @@ while ($true) {
...
         $uuid = $task.content -replace ".*?([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}).*", '$1'
         $auth = $task.content -replace '^[^/]*//([^/]*)/.*', '$1'
 
         if ($uuid -and $auth -eq $ca) {
             Invoke-WebRequest -Uri $task.content -OutFile (New-Item -path "$dir\$uuid.exe" -Force ) -UseBasicParsing
-            $code = Execute "$dir\$uuid.exe"
+            $stdoutTempFile = New-TemporaryFile
+            $stderrTempFile = New-TemporaryFile
+            $code = Execute "$dir\$uuid.exe" $stdoutTempFile $stderrTempFile
+            $stdout = Get-Content -Path $stdoutTempFile
```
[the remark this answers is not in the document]
keitharobertson: I could `Write-Host` and that would write directly to stdout instead of returning, but I wasn't sure if that had any implications for the calling service and I don't have a great way to test that before shipping so I went with what I knew would work.

shell/probe/raindrop.ps1
keitharobertson: sad day
```
Start-Process : This command cannot be run because "RedirectStandardOutput" and "RedirectStandardError" are same. Give
different inputs and Run your command again.
```

shell/probe/raindrop.ps1
makk94: @keitharobertson is there a change accompanying this thumbs up emoji ha

but also, can we just remove these two lines and wait until the test cycle is complete and $dir itself is deleted?

keitharobertson: oops 🤦 I wrote the code, but then accidentally reverted it. But yes you're right I can just let them get deleted with dir.

shell/probe/raindrop.ps1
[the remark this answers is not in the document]
keitharobertson: doing some testing on this to determine what is possible.

shell/probe/raindrop.ps1
keitharobertson: I got rid of these completely. Will get deleted with the dir deletion

shell/probe/raindrop.ps1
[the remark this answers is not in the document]
keitharobertson: `Write-Host` works fine so I moved these into exec

shell/probe/raindrop.ps1
```diff
@@ -31,15 +40,15 @@ while ($true) {
...
         } -UseBasicParsing -MaximumRedirection 0 -ErrorAction SilentlyContinue
-        
+
         $uuid = $task.content -replace ".*?([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}).*", '$1'
         $auth = $task.content -replace '^[^/]*//([^/]*)/.*', '$1'
 
         if ($uuid -and $auth -eq $ca) {
             Invoke-WebRequest -Uri $task.content -OutFile (New-Item -path "$dir\$uuid.exe" -Force ) -UseBasicParsing
-            $code = Execute "$dir\$uuid.exe"
+            $code = Execute "$dir\$uuid.exe" $stdoutTempFile $stderrTempFile
```
makk94, line 51: extra params?

shell/probe/raindrop.ps1
```diff
@@ -1,11 +1,20 @@
...
+        $stdout = Get-Content -Path $stdoutTempFile
+        $stdout = if ($stdout) { [string]::Join("; ", $stdout) } else { "" }
+        $stderr = Get-Content -Path $stderrTempFile
+        $stderr = if ($stderr) { [string]::Join("; ", $stderr) } else { "" }
+        if ($stdout -or $stderr) {
+            Write-Host "${stdout}; ${stderr}"
+        }
         return $Code
     } catch [System.UnauthorizedAccessException] {
         return 126
```
makk94, line 20: will things still get written if you end up in here? (/ did they ever?)

shell/probe/raindrop.ps1
keitharobertson: oops good catch! Problem with testing in an separate environment from my code and copying the changes back 😞

shell/probe/raindrop.ps1
keitharobertson: oh no they will not... I'm going to stick the log stuff in a finally block instead

makk94 (approved):

PR #1016, opened 2024-05-15
DET-1578: windows probe kills tests after 45

packplusplus:
Pids were waiting forever. This may kill tests but not children of children of tests. It seems like a fair trade off. Also, some consistency fixes for var names.

### DIFF
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -9,7 +9,7 @@ function Execute {
         $proc = Start-Process -FilePath $file -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile
 
         $timeouted = $null
-        $proc | Wait-Process -Timeout 60 -ErrorAction SilentlyContinue -ErrorVariable timeouted
+        $proc | Wait-Process -Timeout 45 -ErrorAction SilentlyContinue -ErrorVariable timeouted
 
         if ($timeouted)
         {
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -12,6 +12,7 @@ function Execute {
         $proc | Wait-Process -Timeout 45 -ErrorAction SilentlyContinue -ErrorVariable errVar
 
         if ($errVar) {
+            Write-Host $errVar
             $proc | kill
             return 102
         }
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -11,8 +11,7 @@ function Execute {
         $timeouted = $null
         $proc | Wait-Process -Timeout 45 -ErrorAction SilentlyContinue -ErrorVariable timeouted
 
-        if ($timeouted)
-        {
+        if ($timeouted) {
             $proc | kill
             return 102
         }
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -1,12 +1,12 @@
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 
 function Execute {
-    Param([String]$file)
+    Param([String]$File)
 
     try {
         $stdoutTempFile = New-Item -path "$dir\stdout.log" -Force
         $stderrTempFile = New-Item -path "$dir\stderr.log" -Force
-        $proc = Start-Process -FilePath $file -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile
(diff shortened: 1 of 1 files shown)


shell/probe/raindrop.ps1
```diff
@@ -1,14 +1,24 @@
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 
 function Execute {
-    Param([String]$File)
+    Param([String]$file)
```
keitharobertson, line 4: https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/strongly-encouraged-development-guidelines?view=powershell-7.4#use-pascal-case-for-parameter-names

I don't make the rules 🤷

shell/probe/raindrop.ps1
```diff
@@ -1,14 +1,24 @@
...
+        $proc = Start-Process -FilePath $file -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile
+
+        $timeouted = $null
+        $proc | Wait-Process -Timeout 60 -ErrorAction SilentlyContinue -ErrorVariable timeouted
+
+        if ($timeouted)
+        {
+            $proc | kill
+            return 102
+        }
```
keitharobertson, line 18: nit everywhere else uses this style 
```suggestion
        if ($timeouted) {
            $proc | kill
            return 102
        }
```

keitharobertson (approved):

shell/probe/raindrop.ps1
```diff
@@ -1,14 +1,24 @@
...
 
     try {
         $stdoutTempFile = New-Item -path "$dir\stdout.log" -Force
         $stderrTempFile = New-Item -path "$dir\stderr.log" -Force
-        $R = (Start-Process -FilePath $File -Wait -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile).ExitCode
-        $Code = if (Test-Path $File) {$R} Else {127}
-        return $Code
+        $proc = Start-Process -FilePath $file -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile
+
+        $timeouted = $null
```
makk94, line 11: just b/c this is public code... should we choose a real word ha

shell/probe/raindrop.ps1
```diff
@@ -32,7 +42,7 @@ function FromEnv { param ([string]$envVar, [string]$default)
 }
 
 $ca = FromEnv "PRELUDE_CA" "prelude-account-us1-us-east-2.s3.amazonaws.com"
-$dir = FromEnv "PRELUDE_DIR" $ca
+$dir = FromEnv "PRELUDE_DIR" ".vst"
```
makk94, line 45: the other diff is that nocturnal doesn't even check for an env variable PRELUDE_DIR. do we care about aligning there too?

shell/probe/raindrop.ps1
```diff
@@ -6,9 +6,18 @@ function Execute {
     try {
         $stdoutTempFile = New-Item -path "$dir\stdout.log" -Force
         $stderrTempFile = New-Item -path "$dir\stderr.log" -Force
-        $R = (Start-Process -FilePath $File -Wait -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile).ExitCode
-        $Code = if (Test-Path $File) {$R} Else {127}
-        return $Code
+        $proc = Start-Process -FilePath $File -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile
+
+        $errVar = $null
```
makk94, line 11: nit: `timeoutErr` is clearer

shell/probe/raindrop.ps1
```diff
@@ -6,9 +6,18 @@ function Execute {
...
+        $proc | Wait-Process -Timeout 45 -ErrorAction SilentlyContinue -ErrorVariable errVar
+
+        if ($errVar) {
+            $proc | kill
```
makk94, line 15: curious if you tested / know if the early kill here affects the redirected stdout/err at all?

keitharobertson (approved):

shell/probe/raindrop.ps1
packplusplus: It's on my list to test, but I haven't tested it yet.

makk94 (approved):

shell/probe/raindrop.ps1
packplusplus: I will acquiesce to that once we are sure you can't get messages that aren't timeouts.

PR #1017, opened 2024-05-15
DET-1578: bash probe kills tests after 45

packplusplus:


shell/probe/nocturnal.sh
```diff
@@ -11,7 +11,20 @@ do
 
     if [ "$uuid" ] && [ "$auth" = "$ca" ];then
         curl -sf --max-redirs 0 --create-dirs -o $vst/$uuid $task
-        chmod +x $vst/$uuid && $vst/$uuid; code=$?
+        chmod +x $vst/$uuid && $vst/$uuid & test_pid=$!
+        elapsed_time=0
+        while kill -0 $test_pid 2>/dev/null; do
+          if [ $elapsed_time -ge 60 ]; then
+            kill -9 $test_pid 2> /dev/null
```
makk94, line 18: not sure we should `2>/dev/null`

shell/probe/nocturnal.sh
```diff
@@ -11,7 +11,20 @@ do
 
     if [ "$uuid" ] && [ "$auth" = "$ca" ];then
         curl -sf --max-redirs 0 --create-dirs -o $vst/$uuid $task
-        chmod +x $vst/$uuid && $vst/$uuid; code=$?
+        chmod +x $vst/$uuid && $vst/$uuid & test_pid=$!
+        elapsed_time=0
+        while kill -0 $test_pid 2>/dev/null; do
+          if [ $elapsed_time -ge 60 ]; then
```
makk94, line 17: BE has some logic that assumes tests run within 60s of each other within a test cycle - can we drop the timeout here to 45 or something?

makk94:
just curious, what have you tested this on?

shell/probe/nocturnal.sh
packplusplus: kill -0's exists 0 with no output if a pid exists, when the pid stops existing you get an error like `-bash: kill: (51096) - No such process` which would end up in the error log.

shell/probe/nocturnal.sh
keitharobertson: what logic was that?

makk94: https://github.com/preludeorg/service/blob/main/layers/backbone/backbone/modules/impl/probe.py#L48

shell/probe/nocturnal.sh
makk94: the comment was just about line 18, but if that will basically never error for a "real" reason, then can leave as is

nit then: make the spacing in `2>/dev/null` consistent between lines 16 and 18

shell/probe/nocturnal.sh
```diff
@@ -11,7 +11,20 @@ do
 
     if [ "$uuid" ] && [ "$auth" = "$ca" ];then
         curl -sf --max-redirs 0 --create-dirs -o $vst/$uuid $task
-        chmod +x $vst/$uuid && $vst/$uuid; code=$?
+        chmod +x $vst/$uuid && $vst/$uuid & test_pid=$!
```
keitharobertson, line 14: should this be
```suggestion
        chmod +x $vst/$uuid && $vst/$uuid && test_pid=$!
```

shell/probe/nocturnal.sh
```diff
@@ -11,7 +11,20 @@ do
...
+        elapsed_time=0
+        while kill -0 $test_pid 2>/dev/null; do
+          if [ $elapsed_time -ge 60 ]; then
+            kill -9 $test_pid 2> /dev/null
+            code=102
+          fi
+          sleep 1
+          elapsed_time=$((elapsed_time + 1))
+        done
+        if [[ $code -ne 102 ]]; then
```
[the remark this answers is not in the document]
packplusplus: Yeah, yeah i do. Thank you.

packplusplus:
> just curious, what have you tested this on?

Mostly my mac. I did throw it on an amazon container for giggles, but i was mostly testing hanging tests, not a mix of hanging / not hanging, hence missing the code reset.

shell/probe/nocturnal.sh
```diff
@@ -11,7 +11,20 @@ do
 
     if [ "$uuid" ] && [ "$auth" = "$ca" ];then
         curl -sf --max-redirs 0 --create-dirs -o $vst/$uuid $task
-        chmod +x $vst/$uuid && $vst/$uuid; code=$?
+        chmod +x $vst/$uuid && $vst/$uuid & test_pid=$!
+        elapsed_time=0
+        while kill -0 $test_pid 2>/dev/null; do
```
keitharobertson, line 16: it is unlikely, but technically possible for pid recycling to cause problems here

makk94:
> > just curious, what have you tested this on?
> 
> Mostly my mac. I did throw it on an amazon container for giggles, but i was mostly testing hanging tests, not a mix of hanging / not hanging, hence missing the code reset.

👍 was more curious about platform (mac/al2/ubuntu/what-have-you)

shell/probe/nocturnal.sh
keitharobertson: https://stackoverflow.com/a/22384727
something like this would make it less likely- no sleep for the pid recycle to happen during

packplusplus: I'll take suggestons for other approaches, I'm not sure else to do it. `killall -u ${USER} ${uuid}` has it's own problems.

shell/probe/nocturnal.sh
keitharobertson: oh right yeah it's expecting a response within 60

shell/probe/nocturnal.sh
packplusplus: We chatted in a slack thread about this. Keith and I are convinced we need them. Mahina is defering.

shell/probe/nocturnal.sh
```diff
@@ -11,8 +11,23 @@ do
...
+        done
+        if [[ $code -ne 102 ]]; then
+          wait $test_pid
+          code=$?
+        fi
         dat="${uuid}:$([ -f $vst/$uuid ] && echo $code || echo 127)"
+        unset code
     elif [ "$task" = "stop" ];then
         exit
     else
```
makk94, line 33: while we're sneaking in unrelated changes, i've always been jealous that raindrop gives you a "test cycle done" message before sleeping

makk94 (approved):

shell/probe/nocturnal.sh
keitharobertson: I agree! I never thought I would see the day where Mahina asks for a repetitive log though! I'm going to print this out and frame it.

makk94: lol not _every_ log is repetitive! i had >0 logs before you came on board

PR #1018, opened 2024-05-20
det-1124: nocturnal shouldn't follow redirects

packplusplus:
Smole change for nocturnal.


makk94 (approved):
personally feel we should be deleting the `MaximumRedirection` from raindrop instead of adding it here once we move to returning 200s. b/c extraneous code. but i wont stand in the way of your happiness

PR #1021, opened 2024-05-21
fix/windows probe

packplusplus:
This will output a timeout note. But I'm sure there's debate on the best way to do this. So this is my starting volley.

### DIFF
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -9,7 +9,7 @@ function Execute {
         $proc = Start-Process -FilePath $file -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile
 
         $timeouted = $null
-        $proc | Wait-Process -Timeout 60 -ErrorAction SilentlyContinue -ErrorVariable timeouted
+        $proc | Wait-Process -Timeout 45 -ErrorAction SilentlyContinue -ErrorVariable timeouted
 
         if ($timeouted)
         {
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -12,6 +12,7 @@ function Execute {
         $proc | Wait-Process -Timeout 45 -ErrorAction SilentlyContinue -ErrorVariable errVar
 
         if ($errVar) {
+            Write-Host $errVar
             $proc | kill
             return 102
         }
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -11,8 +11,7 @@ function Execute {
         $timeouted = $null
         $proc | Wait-Process -Timeout 45 -ErrorAction SilentlyContinue -ErrorVariable timeouted
 
-        if ($timeouted)
-        {
+        if ($timeouted) {
             $proc | kill
             return 102
         }
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -1,12 +1,12 @@
 [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
 
 function Execute {
-    Param([String]$file)
+    Param([String]$File)
 
     try {
         $stdoutTempFile = New-Item -path "$dir\stdout.log" -Force
         $stderrTempFile = New-Item -path "$dir\stderr.log" -Force
-        $proc = Start-Process -FilePath $file -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile
+        $proc = Start-Process -FilePath $File -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile
 
         $timeouted = $null
         $proc | Wait-Process -Timeout 45 -ErrorAction SilentlyContinue -ErrorVariable timeouted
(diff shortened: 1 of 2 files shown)


shell/probe/raindrop.ps1
```diff
@@ -9,10 +9,9 @@ function Execute {
         $proc = Start-Process -FilePath $File -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile
 
         $errVar = $null
```
makk94, line 11: update?

shell/probe/raindrop.ps1
```diff
@@ -9,10 +9,9 @@ function Execute {
...
-        $proc | Wait-Process -Timeout 45 -ErrorAction SilentlyContinue -ErrorVariable errVar
+        $proc | Wait-Process -Timeout 10 -ErrorAction SilentlyContinue -ErrorVariable timeoutVar
```
keitharobertson, line 12: 10?

shell/probe/raindrop.ps1
```diff
@@ -30,6 +29,7 @@ function Execute {
         $stdout = if ($stdout) { [string]::Join("; ", $stdout) } else { "" }
         $stderr = Get-Content -Path $stderrTempFile
         $stderr = if ($stderr) { [string]::Join("; ", $stderr) } else { "" }
+        if ($timeoutVar) {$stdErr = [string]::Join(" ", $stderr, "TEST TIMEOUT; ") }
```
keitharobertson, line 32: `stdErr` doesn't look like it is used anywhere and the rest of the logs are `;` separated. How about this
```suggestion
        if ($timeoutVar) {$stderr += "; TEST TIMEOUT") }
```

keitharobertson: or alternatively this and include $timeoutMsg in the `Write-Host` on line 34.
```suggestion
        $timeoutMsg = if ($timeoutVar) { "TEST TIMEOUT" } else { "" }
```

shell/probe/raindrop.ps1
```diff
@@ -8,11 +8,10 @@ function Execute {
         $stderrTempFile = New-Item -path "$dir\stderr.log" -Force
         $proc = Start-Process -FilePath $File -NoNewWindow -PassThru -RedirectStandardOutput $stdoutTempFile -RedirectStandardError $stderrTempFile
 
-        $errVar = $null
-        $proc | Wait-Process -Timeout 45 -ErrorAction SilentlyContinue -ErrorVariable errVar
+        $timeoutVar = $null
```
makk94, line 11: do you even need this line? You said you tested and it worked when this line was still errVar, so maybe powershell defaults undeclared variables to null already?

packplusplus: Yes, it's needed. Also that value exists again. 
```
while (1) {
 $proc = Start-Process -FilePath "notepad.exe" -NoNewWindow -PassThru
 if ($timeoutVar -eq $null) {
  Write-Host "is null"
 } else {
  Write-Host "is not null"
 }
 $proc | Wait-Process -Timeout 5 -ErrorAction SilentlyContinue -ErrorVariable timeoutVar
 Write-Host $timeoutVar
 $proc | kill
}
```

```
PS C:\Users\Administrator\Desktop> .\t.ps1
is null
This command stopped operation because process "notepad (7692)" is not stopped in the specified time-out.
is not null
This command stopped operation because process "notepad (444)" is not stopped in the specified time-out.
is not null
This command stopped operation because process "notepad (7324)" is not stopped in the specified time-out.
is not null
This command stopped operation because process "notepad (424)" is not stopped in the specified time-out.
is not null
This command stopped operation because process "notepad (7424)" is not stopped in the specified time-out.
```

makk94: the while loop is outside this function, and hopefully powershell isnt so crazy as to keep that variable around between Executes

this is small potatoes, i'm just looking for ways to cut down chars

shell/probe/raindrop.ps1
packplusplus: I think I like the approach I took. Thoughts?

keitharobertson (approved):

makk94 (approved):

shell/probe/raindrop.ps1
keitharobertson: looks good to me

shell/probe/raindrop.ps1
packplusplus: Yeah, you're right 
```
function Execute {
 $proc = Start-Process -FilePath "notepad.exe" -NoNewWindow -PassThru
 if ($timeoutVar -eq $null) {
  Write-Host "is null"
 } else {
  Write-Host "is not null"
 }
 $proc | Wait-Process -Timeout 5 -ErrorAction SilentlyContinue -ErrorVariable timeoutVar
 Write-Host $timeoutVar
 $proc | kill
}


while (1) {
 Execute
}
```

```
.\t.ps1                                                                              is null                                                                                                                 This command stopped operation because process "notepad (5840)" is not stopped in the specified time-out.               is null                                                                                                                 This command stopped operation because process "notepad (6164)" is not stopped in the specified time-out.               is null                                                                                                                 This command stopped operation because process "notepad (6072)" is not stopped in the specified time-out.               is null
This command stopped operation because process "notepad (10232)" is not stopped in the specified time-out.
```

PR #1022, opened 2024-05-22
cleanup: log missing Exitcode as string

makk94:
In error cases, we get empty string exit codes, so this is just to avoid the uncaught exception of
```
TypeError: %d format: a real number is required, not str
```


packplusplus (approved):
👍

PR #1031, opened 2024-06-03
fix py sdk tests: new compile test output

makk94:


keitharobertson (approved):
nice! One of these days I will remember to update these tests

PR #1033, opened 2024-06-05
DET-1611: Add SDK and CLI functionality for threat hunt queries

keitharobertson:


python/cli/prelude_cli/views/build.py
```diff
@@ -270,3 +270,53 @@ def delete_detection(controller, detection):
     with Spinner(description='Removing detection'):
         data = controller.delete_detection(detection_id=detection)
     print_json(data=data)
+
+
+@build.command('create-threat-hunt')
+@click.option('-n', '--name', help='Name of this threat hunt query', required=True, type=str)
```
makk94, line 276: name is an arg for our other resources

python/sdk/tests/test_build.py
```diff
@@ -246,4 +246,58 @@ def test_update_detection(self, unwrap):
     def test_delete_detection(self, unwrap):
         unwrap(self.build.delete_detection)(self.build, detection_id=pytest.detection_id)
         with pytest.raises(Exception):
-            unwrap(self.detect.get_detection)(self.detect, detection_id=pytest.detection_id)
\ No newline at end of file
+            unwrap(self.detect.get_detection)(self.detect, detection_id=pytest.detection_id)
+
+@pytest.mark.order(4)
```
makk94, line 251: intentional to use the same order index as the Detections tests? don't think the two should conflict, just  checking intention

python/sdk/tests/test_build.py
```diff
@@ -246,4 +246,58 @@ def test_update_detection(self, unwrap):
...
+        diffs = check_dict_items(pytest.expected_threat_hunt, mine[0])
+        assert not diffs, json.dumps(diffs, indent=2)
+
+    def test_update_threat_hunt(self, unwrap):
+        res = unwrap(self.build.update_threat_hunt)(
+            self.build,
+            threat_hunt_id=pytest.threat_hunt_id,
+            name='updated threat hunt',
+            query='updated query')
+        assert res['name'] == 'updated threat hunt'
```
makk94, line 296: will you update pytest.expected_threat_hunt? i know it's not used anywhere later right now, but as a principle, I'd like the pytest.expected_* stuff to always be correct

makk94 (approved):

python/sdk/tests/test_build.py
makk94: i have. but maybe that just means i'm a control freak. i'm fine with leaving this as is

makk94: also, practically, when they have the same order index, i think they'll just run in order of the file. so detections will always run first anyway

keitharobertson: 😆 makes no difference to me so I'm happy to increment

makk94: if you do, you have to increment al the ones that come after it. and then probably also the order indexes on the delete test (last test of this lil module) as well.

makk94: ^ prefer you either leave the order index the same as the Detection tests or you update all the other ones in the other files too (test_detect, test_partner, test_probe, test_generate) so we're unique across the board

PR #1035, opened 2024-06-10
DET-1638: Update endpoint.go

smprelude:
In this PR:
- Pruned a couple of exit codes that we don't use from the endpoint library
-- OutOfMemory (137) - not used, use cases uncertain/absent
-- TestIncorrectlyBlocked (110) - not used, use cases absent
- Also corrected spelling of 'HostNotVulnerabile', will submit separate PR in tests repo to correct any VSTs which emit this code

### DIFF
--- a/go/tests/endpoint/endpoint.go
+++ b/go/tests/endpoint/endpoint.go
@@ -22,7 +22,6 @@ const (
 	UnexpectedTestError      int = 1
 	TimeoutExceeded          int = 102
 	CleanupFailed            int = 103
-	OutOfMemory              int = 137
 	UnexpectedExecutionError int = 256
 
 	// Not Relevant
@@ -32,13 +31,12 @@ const (
 	TestCompletedNormally       int = 100
 	FileQuarantinedOnExtraction int = 105
 	NetworkConnectionBlocked    int = 106
-	HostNotVulnerabile          int = 107
+	HostNotVulnerable           int = 107
 	ExecutionPrevented          int = 126
 	FileQuarantinedOnExecution  int = 127
 
 	// Unprotected
 	Unprotected            int = 101
-	TestIncorrectlyBlocked int = 110
 )
 
 type fn func()


makk94:
@packplusplus so when this library change goes in, compute will immediately (minus whatever warm lambdas) start pulling this new code, right? So this change should go in only when the VST update is also ready to go to us1?

packplusplus:
> @packplusplus so when this library change goes in, compute will immediately (minus whatever warm lambdas) start pulling this new code, right? So this change should go in only when the VST update is also ready to go to us1?

That's how I understand it. Maybe we do some transition plan for the change? so we can have the fix in. The translation only goes one way right? So people could fix misspellings.

Would we have to look for tests that may reference the wrong spelling? 

Is the wrong spelling in other libs? Do you want to talk about how 107 is EXPLOIT_PREVENTED in the python codes?

smprelude:
the three tests that you found in tests/* have been addressed in a separate pr to the tests repo. vsts that exit with the int instead of the enum'd string won't be affected by the spell change.

i haven't located any other usage of the misspelling, at least in tests/*

one way to do this would be to make the change in tests/* first (so the vsts affected by this change are amended), merge that in, THEN merge this change in (pending approval) once the tests are propagated out. this order works because test promo from repo->us2->us1/eu1 takes more time than the lib changes would.

thoughts?

makk94:
A multi-stage deploy could also be:
1. Add the correct spelling but leave the bad spelling also
2. Deploy the VSTs with the fixed spelling
3. Remove the bad spelling 

All that if we're comfortable assuming no custom tests are using the bad spelling

bfuzzy1 (approved):

smprelude:
talked with james today. we agreed to kill PRs relating to changing endpoint.go until we solidify a procedure for making changes to the library. keith and i (and other stakeholders) will work on that.

PR #1040, opened 2024-06-18
fix/update threat intel CLI

makk94:


python/cli/prelude_cli/views/build.py
```diff
@@ -293,9 +309,9 @@ def create_threat_hunt(controller, name, query, test, threat_hunt_id):
 
 @build.command('update-threat-hunt')
 @click.argument('threat_hunt')
-@click.option('-n', '--name', help='Name of this threat hunt query', required=True, type=str)
-@click.option('-q', '--query', help='Threat hunt query', required=True, type=str)
-@click.option('-t', '--test', help='ID of the test this threat hunt query is for', required=True, type=str)
+@click.option('-n', '--name', help='Name of this threat hunt query', type=str)
+@click.option('-q', '--query', help='Threat hunt query', type=str)
+@click.option('-t', '--test', help='ID of the test this threat hunt query is for', type=str)
```
keitharobertson, line 314: are these not required?

python/cli/prelude_cli/views/generate.py
```diff
@@ -34,11 +34,15 @@ def generate_threat_intel(controller: GenerateController, threat_pdf: str, outpu
...
                 with open(f'{output_dir}/{technique_directory}/test.go', 'w') as f:
-                    f.write(technique['go_code'])
-                for i, sigma_rule in enumerate(technique['sigma_rules']):
+                    f.write(content['go_code'])
+                for i, sigma_rule in enumerate(content['sigma_rules']):
                     with open(f'{output_dir}/{technique_directory}/sigma_{i}.yaml', 'w') as f:
                         f.write(sigma_rule)
+                for i, query in enumerate(content.get('threat_hunt_queries', [])):
+                    with open(f'{output_dir}/{technique_directory}/query_{i}.txt', 'w') as f:
+                        f.write(query['query'])
```
keitharobertson, line 45: I think this should be a json file and write the query object including the name. Need the name to create it later.

python/cli/prelude_cli/views/build.py
```diff
@@ -149,25 +149,35 @@ def upload(p: Path):
...
+                        go_code = f.read()
+                        test_uploads.append(controller.upload(test_id=test['id'], filename=f'{test["id"]}.go', data=go_code.encode()))
+                    for sigma_file in Path(f'{directory}/{technique_dir}').glob('sigma*'):
+                        with open(sigma_file, 'r') as f:
+                            rule = f.read()
+                            created_detections.append(controller.create_detection(rule=rule, test_id=test['id']))
+                    for query_file in Path(f'{directory}/{technique_dir}').glob('query*'):
+                        with open(query_file, 'r') as f:
+                            queries = f.read()
+                            created_queries.append(controller.create_threat_hunt(name=config['name'], query=queries, test_id=test['id']))
```
keitharobertson, line 180: each query should have a specific name- not the test name

python/cli/prelude_cli/views/build.py
makk94: in an update, users should only set the fields they want to change - if they want to leave the value as is, they dont set it

python/cli/prelude_cli/views/generate.py
makk94: done

python/cli/prelude_cli/views/build.py
makk94: done

keitharobertson (approved):

PR #1047, opened 2024-07-17
Endpoint Library IPC Functionality

smprelude:
In this PR:
- we implement a dropper execution flow to be used in conjunction with VSTs which write malicious files to disk at time-of-test. this dropper is employed as a sacrificial child process of the vst process. it is implemented because we observed AV/EDR interdicting at time-of-test on those tests which write to the FS by killing the vst process, which would short-circuit the vst's code path, preventing them from emitting the proper exit code -- for example, instead of exiting PROTECTED::FileQuarantinedOnExtraction (int 127), we would instead see exit codes like (int -1). the execution model we adopt with this dropper program is such that the AV/EDR interdiction will target the dropper and not the vst process, allowing for the vst to emit the proper exit code and thus preserve test result value for the customer base.

- the dropper operates by opening a file socket on the host in the testing directory. the vst and the dropper communicate via this socket. the dropper "listens" on the file socket, and then writes gob-encoded data it receives through the socket to the victim file system.
-- adopting this model, execution flow for relevant tests will look like this:
1) VSTs now embed dropper program (not all of them, just the ones that do this)
2) VST writes embedded dropper executable to disk
3) VST starts the dropper as a child of itself
4) dropper removes any instances of the prelude socket file ("prelude_socket") from testing directory
5) dropper writes new socket file to testing directory and starts to "listen" on it
6) VST connects to the file socket
7) VST gob-encodes its payload (the malicious file) and sends it over the file socket (writes to socket file)
8) dropper receives gob-encoded payload, reads it in, decodes it, and writes the malicious file to the path specified in the gob
9) VST process does the rest of the test as they currently are encoded from here

- implemented DropperPayload data structure
-- this struct has two members: filename(str) and contents([]byte). these are set by the VST process at time-of-test.

- implemented socketPath global variable in endpoint library. this is used both to specify the path to the file socket, as well as a bat signal by Endpoint::Write() to know whether or not the dropper execution model will be employed
-- socketPath is an encapsulated (private, non-static) variable not exported by the endpoint library for safety & stability purposes
-- to set/clear, we implement mutator functions SetSocketPath() and ClearSocketPath()
-- we don't implement an accessor for the socketPath field because we want that to be invariant (always "prelude_socket").

- implemented writeIPC() function into the endpoint library
-- this function is responsible for carrying out writes to the socket. it is invoked by Write() when Endpoint.SocketPath is not emptystring.
-- writeIPC() is a private library function which: starts the dropper child process, connects to the file socket, encodes the payload, writes to the socket, and then performs a failsafe dropper process kill.

- implemented startDropperChildProcess() as a private library function which: switches on host platform to build the command line string used to invoke the dropper program; this varies based on whether or not the host OS is windows. it returns a golang process object and an error (hProc and nil upon success).
-- startDropperChildProcess() is invoked by writeIPC() at the top of the function; if we can't start the dropper then we can't do IPC-enabled file writing

- Write() and writeIPC() now return errors. when consumed, these errors provide valuable insight into why a write op failed. Quarantined() consumes these errors.

- Endpoint::Quarantined() now uses Endpoint.Exists() to perform its file check

- alphabetized function name order in the endpoint library :')

### DIFF
--- a/go/tests/endpoint/endpoint.go
+++ b/go/tests/endpoint/endpoint.go
@@ -5,10 +5,12 @@ import (
 	"bytes"
 	"crypto/aes"
 	"crypto/cipher"
+	"encoding/gob"
 	"fmt"
 	"io"
 	"io/fs"
 	"math/rand"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


go/tests/endpoint/endpoint.go
```diff
@@ -197,6 +173,18 @@ func IsAvailable(programs ...string) bool {
 	return false
 }
 
+func IsSecure() bool {
```
bfuzzy1, line 176: Why is this back?

smprelude: copy/paste error on my part. i thought it was weird that it was back :)

bfuzzy1 (commented):
@smprelude any screenshots from testing? What does a VST look like that uses this change in raw golang?

smprelude:
> @smprelude any screenshots from testing? What does a VST look like that uses this change in raw golang?
here are a couple screenshots: one for usage in VST code, one for output
image
image

garjarbinks:
> note that usage in vst writing is pretty straightforward; just call setsocketpath() to throw up the signal so that write() calls writeipc(), then after the quarantined() call proceed as normal (assuming a file socket is there to connect/write to)

as customers are using our libs, should this be named something more universal?
e.g. `set_subprocess_write()` or something equally boring?
or, considering we could expand beyond write:
`set_subprocess_crud()`

garjarbinks:
QQ: where is the prelude_dropper? are we including that in our library too?

smprelude:
> > note that usage in vst writing is pretty straightforward; just call setsocketpath() to throw up the signal so that write() calls writeipc(), then after the quarantined() call proceed as normal (assuming a file socket is there to connect/write to)
> 
> as customers are using our libs, should this be named something more universal? e.g. `set_subprocess_write()` or something equally boring? or, considering we could expand beyond write: `set_subprocess_crud()`

hm. we could rename it to something that changes the perspective of intent on the caller's part. what about "DoWriteAsChild"? i agree that "SetSocketPath" might too heavily imply that the user can set the string despite it being an invariant.

smprelude:
> QQ: where is the prelude_dropper? are we including that in our library too?

technically the dropper source code isn't a part of the endpoint library or the libraries repo so that's why it wasn't published. maybe it should be, though!

bfuzzy1:
> > QQ: where is the prelude_dropper? are we including that in our library too?
> 
> technically the dropper source code isn't a part of the endpoint library or the libraries repo so that's why it wasn't published. maybe it should be, though!

likely to get signatured faster if it is public if that is of any concern for us.

smprelude:
> likely to get signatured faster if it is public if that is of any concern for us.
> 
> EDIT: i see no issue as customers will need to understand the dropper a bit deeper and should have a general Prelude public dropper they can use with their own custom tests worth bringing up just due to the fact of AV.

fwiw i agree. for any customer-written tests, they just need to code their own dropper or ensure a unix dom socket is available in the testing directory at runtime.

go/tests/endpoint/endpoint.go
```diff
@@ -207,82 +183,105 @@ func Pwd(filename string) string {
...
-	if len(data) < nonceSize {
-		return nil, fmt.Errorf("ciphertext too short")
+func startDropperChildProcess() (*os.Process, error) {
+	execDir := filepath.Dir(socketPath)
+
+	var listenerPath string
+
+	switch platform := GetOS(); platform {
+	case "windows":
+		listenerPath = filepath.Join(execDir, "prelude_dropper.exe")
```
packplusplus, line 260: Can these be named dynamically `$test_id_dropper` so it's easier to correlate events related to the pid if relationships are lost, or if the pids don't get cleaned up for some reason?

go/tests/endpoint/endpoint.go
```diff
@@ -319,3 +318,82 @@ func Unzip(zipData []byte) error {
...
+	} else {
+		Say(fmt.Sprintf("Waiting for %d seconds", dur))
+		time.Sleep(dur * time.Second)
+	}
+}
+
+func Write(filename string, contents []byte) error {
+	var err error
+	if socketPath != "" {
+		err = writeIPC(filename, contents)
```
packplusplus, line 336: can we log which write we use here so we better understand the code path a test took?

go/tests/endpoint/endpoint.go
smprelude: done in new commit. good call!!

go/tests/endpoint/endpoint.go
smprelude: tried to address this in new commit with a Say() call indicating style of write to be performed just before the operation is performed

smprelude:
screenie with new increased logging and dynamic dropper naming
image

smprelude:
latest:
- adjusted the Exists() call in Quarantined() to use literal path as discerned from PRELUDE_DIR path variable
- Write() now also does this
- modified Say() to function as a format printer. this allows future tests to directly embed format strings without directly invoking fmt into test source, which can be useful for writing out error info, command output, and more

bfuzzy1 (approved):

PR #1061, opened 2024-08-07
add new exit code to sdk (INSUFFICIENT_PRIVILEGES 109)

bfuzzy1:


makk94:
Did y'all want to add back 110 too?

makk94:
Also would you add to our docs as well https://docs.preludesecurity.com/docs/understanding-results

bfuzzy1:
> Did y'all want to add back 110 too?

cc'ing @smprelude on this one. I don't think we should i can make the pr to remove it from endpoint if Scott agree's. if not we should discuss re-definition as its defined as being health check related really.

smprelude:
let's not necro 110. health-check style tests aren't in active production and so there won't be any reason to have it save for backwards compatibility with old test assets that utilize it.

bfuzzy1:
> let's not necro 110. health-check style tests aren't in active production and so there won't be any reason to have it save for backwards compatibility with old test assets that utilize it.

After some discussion. We should have it on until we refresh / update the agent potentially so I will add it back.

We do have tests that use it currently.

python/sdk/prelude_sdk/models/codes.py
```diff
@@ -112,11 +114,13 @@ def mapping(cls):
                 ExitCode.UNEXPECTED_ERROR,
                 ExitCode.UNKNOWN_ERROR,
                 ExitCode.UNREPORTED,
+                ExitCode.INCORRECTLY_BLOCKED,
```
makk94, line 117: nit: alphabetize these 2 error state mappings

makk94 (approved):

PR #1062, opened 2024-08-13
fix sdk tests; add new APIs (to CLI / SDK / tests)

makk94:
added or updated:
- compile code string
- do threat hunt
- threat hunt activity
- ioa stats
- observed detected
- list partner advisories
- generate from partner advisory
- undelete test
- undelete threat
- list techniques
- partner reports
- partner block
- get test
- list tests

### DIFF
--- a/python/cli/prelude_cli/views/partner.py
+++ b/python/cli/prelude_cli/views/partner.py
@@ -128,3 +128,13 @@ def observed_detected(controller, test_id, hours):
     with Spinner(description='Getting observed / detected stats'):
         data = controller.observed_detected(test_id=test_id, hours=hours)
     print_json(data=data)
+
+@partner.command('advisories')
+@click.argument('partner', type=click.Choice([Control.CROWDSTRIKE.name], case_sensitive=False))
+@click.pass_obj
+@handle_api_error
+def partner_advisories(controller, partner):
+    """ Get advisories provided by partner """
+    with Spinner(description='Getting partner advisories'):
+        data = controller.list_advisories(partner=Control[partner])
+    print_json(data=data)
--- a/python/sdk/prelude_sdk/controllers/build_controller.py
+++ b/python/sdk/prelude_sdk/controllers/build_controller.py
@@ -1,5 +1,3 @@
-import requests
-
 from prelude_sdk.controllers.http_controller import HttpController
 
 from prelude_sdk.models.account import verify_credentials
@@ -88,6 +86,19 @@ class BuildController(HttpController):
             f'{self.account.hq}/build/tests/{test_id}/{filename}',
             data=data,
             headers=h,
+            timeout=10
+        )
+        if res.status_code == 200:
+            return res.json()
+        raise Exception(res.text)
+
+    @verify_credentials
+    def compile_code_string(self, code: str):
+        """ Compile a code string """
+        res = self._session.post(
+            f'{self.account.hq}/build/compile',
+            json=dict(code=code),
+            headers=self.account.headers,
             timeout=10
         )
         if res.status_code == 200:
--- a/python/sdk/prelude_sdk/controllers/partner_controller.py
+++ b/python/sdk/prelude_sdk/controllers/partner_controller.py
@@ -1,5 +1,4 @@
 from datetime import datetime, timezone
-import requests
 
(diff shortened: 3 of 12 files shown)


python/cli/prelude_cli/views/partner.py
```diff
@@ -128,3 +128,13 @@ def observed_detected(controller, test_id, hours):
...
+
+@partner.command('advisories')
+@click.argument('partner', type=click.Choice([Control.CROWDSTRIKE.name], case_sensitive=False))
+@click.pass_obj
+@handle_api_error
+def partner_advisories(controller, partner):
+    """ Get advisories provided by partner """
+    with Spinner(description='Getting partner advisories'):
+        data = controller.list_advisories(partner=Control[partner])
+    print_json(data=data)
```
keitharobertson, line 140: do we still want to review with FE what fields we should keep before shipping this or just plan to adjust later?

python/sdk/prelude_sdk/controllers/build_controller.py
```diff
@@ -94,6 +92,19 @@ def upload(self, test_id, filename, data):
...
+        """ Compile a code string """
+        res = self._session.post(
+            f'{self.account.hq}/build/compile',
+            json=dict(code=code),
+            headers=self.account.headers,
+            timeout=10
+        )
+        if res.status_code == 200:
+            return res.json()
+        raise Exception(res.text)
```
keitharobertson, line 106: is there reason to add things like this to the public sdk that we don't add to the cli? This doesn't provide much use to anyone outside of the UI- if writing a test, build it locally and then upload to a prelude test and you'll get the compile output there.

python/sdk/prelude_sdk/controllers/partner_controller.py
```diff
@@ -139,3 +138,15 @@ def observed_detected(self, test_id: str | None, hours: int | None):
         if res.status_code == 200:
             return res.json()
         raise Exception(res.text)
+
+    @verify_credentials
+    def list_advisories(self, partner: Control):
+        """ Get advisory reports provided by a partner """
+        res = self._session.get(
+            f'{self.account.hq}/partner/advisories/{partner.name}',
```
keitharobertson, line 146: isn't this the value not the name?

python/cli/prelude_cli/views/partner.py
makk94: since shipping seems imminent, i'm wondering if we should prune now, down just to the ID and then add things back in once FE (and honestly probably Pete / UX) know what they want from it

even if we leave the API out of the CLI / SDK, it's still _technically_ backwards incompatible to change the API later since its a public endpoint

python/sdk/prelude_sdk/controllers/build_controller.py
makk94: yeah idk. it's nice to have an official e2e test of it. and maybe someone exists out there who writes go code but can't / doesnt want to compile locally... lol? pretty weak arguments though, and happy to remove if you think i should

python/sdk/prelude_sdk/controllers/partner_controller.py
makk94: can be either. name is friendlier so it's usually the one i test

python/cli/prelude_cli/views/partner.py
keitharobertson: I like that idea- prune down to the bare minimum and add back as needed. Not a breaking change!

python/sdk/prelude_sdk/controllers/build_controller.py
keitharobertson: the testing argument has me sold. Carry on!

python/sdk/prelude_sdk/controllers/partner_controller.py
keitharobertson: oh right I keep forgetting those enums are magic

keitharobertson (approved):

python/cli/prelude_cli/views/partner.py
makk94: https://github.com/preludeorg/service/pull/2095

PR #1063, opened 2024-08-14
DET-1867: Clearer logging

packplusplus:
Logging was clear in nocturnal, it was only powershell that needed to be updated.

### DIFF
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -18,10 +18,10 @@ function Execute {
         $code = if (Test-Path $File) {$proc.ExitCode} Else {127}
         return $code
     } catch [System.UnauthorizedAccessException] {
-        Write-Host "System.UnauthorizedAccessException - $File"
+        "System.UnauthorizedAccessException - $File" | Out-File -FilePath $stderrTempFile -Append
         return 126
     } catch [System.InvalidOperationException] {
-        Write-Host "System.InvalidOperationException - $File"
+        "System.InvalidOperationException - $File" | Out-File -FilePath $stderrTempFile -Append
         return 127
     } catch {
         return 1
--- a/shell/probe/raindrop.ps1
+++ b/shell/probe/raindrop.ps1
@@ -18,8 +18,10 @@ function Execute {
         $code = if (Test-Path $File) {$proc.ExitCode} Else {127}
         return $code
     } catch [System.UnauthorizedAccessException] {
+        Write-Host "System.UnauthorizedAccessException - $File"
         return 126
     } catch [System.InvalidOperationException] {
+        Write-Host "System.InvalidOperationException - $File"
         return 127
     } catch {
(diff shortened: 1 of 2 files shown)


shell/probe/raindrop.ps1
```diff
@@ -18,8 +18,10 @@ function Execute {
         $code = if (Test-Path $File) {$proc.ExitCode} Else {127}
         return $code
     } catch [System.UnauthorizedAccessException] {
+        Write-Host "System.UnauthorizedAccessException - $File"
         return 126
     } catch [System.InvalidOperationException] {
+        Write-Host "System.InvalidOperationException - $File"
```
keitharobertson, line 24: should these append to the stdout file instead to keep it consistent?

packplusplus: This is an excellent question / point. Let me try that. My other question was with the other 127, and if we should write something out there as well (but maybe that's a better other PR, since I'd also like to tackle the "if the file exists, we should have an exit code" logic)

packplusplus: The output would look like

```
[2024-08-15T11:08:48][359686a3-1181-41b3-a006-2cc1927c9386] Starting test at: 2024-08-15T11:08:48; [2024-08-15T11:08:48][359686a3-1181-41b3-a006-2cc1927c9386] Performing relevancy check; [2024-08-15T11:08:48][359686a3-1181-41b3-a006-2cc1927c9386] Determining if endpoint is conected to Active Directory; [2024-08-15T11:08:49][359686a3-1181-41b3-a006-2cc1927c9386] Completed with code: 104; [2024-08-15T11:08:49][359686a3-1181-41b3-a006-2cc1927c9386] Exit called from line: 37; [2024-08-15T11:08:49][359686a3-1181-41b3-a006-2cc1927c9386] Ending test at: 2024-08-15T11:08:49; ;
; System.InvalidOperationException - C:\Program Files\Prelude Security\Prelude Probe\.vst\6d955e3c-37b4-4327-a811-5a72fd4c1198.exe;
; System.InvalidOperationException - C:\Program Files\Prelude Security\Prelude Probe\.vst\a2725950-6c54-4cae-9a1b-67bf098e70e8.exe;
; System.InvalidOperationException - C:\Program Files\Prelude Security\Prelude Probe\.vst\d7aba28b-b4c9-47e6-8163-9408e7e443af.exe;
``` 

Which I _think_ I like. Anyone parsing the event logs would still be able to alert on "error output".

makk94: are those new lines? are they gonna create separate entries in the event viewer?

packplusplus: > are those new lines? are they gonna create separate entries in the event viewer?

they should. it did in my test env, but until the whole installer is packaged up and pulling them, i won't believe it.

makk94 (approved):

PR #1067, opened 2024-08-21
[DET-1881] Enable build and publish of CLI and SDK via GitHub Actions.

ohlol:


packplusplus:
This needs to notify a slack channel when it does work. It also seems like all the version and git tagging would still be done by hand, is that true? Why not get that automated as well?

ohlol:
> This needs to notify a slack channel when it does work.

ohh yea good point, will add that.

> This needs to notify a slack channel when it does work. It also seems like all the version and git tagging would still be done by hand, is that true? Why not get that automated as well?

we don't have automation for tags for anything else, that's the only reason. we can add it but should probably do it for everything. can look into it as a follow up

.github/workflows/release.yml
```diff
@@ -46,6 +46,25 @@ jobs:
         uses: pypa/gh-action-pypi-publish@@release/v1
         with:
           packages-dir: cli-dist
+      - name: Notify Slack
+        uses: slackapi/slack-github-action@v1.26.0
```
packplusplus, line 50: Needs an `if: always()` so we get failures.

packplusplus:
I think we want to split these out into separate workflows (or one workflow with a single sdk OR cli option). The sdk needs to finish before the cli if they both rev their versions. I think it'd be tempting to rev both, then kick off one job and then it's a race which may sometimes fail. 

I don't know if I prefer one over the other. I guess maybe the or version would be easier to maintain. 

Other than that, yeah, the job looks right, so either split them into two files or make it run in sdk or cli context.

packplusplus (approved):
HOORAY!

PR #1072, opened 2024-09-03
DET-1913/threat_hunt_id_to_id

packplusplus:


python/cli/prelude_cli/views/build.py
```diff
@@ -343,7 +343,7 @@ def update_threat_hunt(controller, threat_hunt, name, query, test):
             name=name,
             query=query,
             test_id=test,
-            threat_hunt_id=threat_hunt,
+            threat_hunt_id=threat_hunt
```
keitharobertson, line 346: super nitty nit: I personally prefer trailing `,` because it makes future PR diffs smaller, but we don't really have a standard for it

python/sdk/tests/test_build.py
```diff
@@ -326,17 +326,17 @@ def test_list_threat_hunts(self, unwrap):
...
         assert not diffs, json.dumps(diffs, indent=2)
 
     def test_update_threat_hunt(self, unwrap):
         pytest.expected_threat_hunt = unwrap(self.build.update_threat_hunt)(
             self.build,
-            threat_hunt_id=pytest.threat_hunt_id,
             name='updated threat hunt',
-            query='.*')
+            query='.*',
+            threat_hunt_id=pytest.threat_hunt_id)
```
keitharobertson, line 339: shouldn't this be `id`?

python/cli/prelude_cli/views/build.py
makk94: it's just feels so unfinished ha but ok

python/sdk/tests/test_build.py
makk94: all the other functions in our SDK (createTest/Threat, etc) name params this explicitly, so sticking with that

python/cli/prelude_cli/views/build.py
keitharobertson: I don't have strong enough feelings to say we have to do it. I just like it when diffs are one line instead of two. We can standardize the other way if you have opinions

keitharobertson (approved):

python/cli/prelude_cli/views/build.py
makk94: i'll try to get in the habit of the trailing comma for multiline lists / params

PR #1073, opened 2024-09-03
DET-1927 - validate nocturnal changes

packplusplus:
See ticket, but the nocturnal code was just broken. Tested on ubuntu arm and x86 dashes, as well as bash 3 and 4 on mac.

### DIFF
--- a/shell/probe/nocturnal.sh
+++ b/shell/probe/nocturnal.sh
@@ -1,10 +1,12 @@
 #!/bin/sh
 
 ca=${PRELUDE_CA:-prelude-account-us1-us-east-2.s3.amazonaws.com}
-vst=${PRELUDE_DIR:-.vst}
+vst=${PRELUDE_DIR:-"$PWD/.vst"}
+
 version='2.6'
 
 echo "Prelude probe: version ${version}"
+
 
 while :
 do
@@ -16,11 +18,9 @@ do
         echo "Downloading $uuid"
         curl -sf --max-redirs 0 --create-dirs -o "$vst/$uuid" $task
         chmod +x "$vst/$uuid"
-        CDIR="${PWD}"
-        cd "$vst" &>/dev/null
         echo "Invoking $uuid"
-        $vst/$uuid & test_pid=$!
-        cd "$CDIR" &>/dev/null
+        cd "$vst" || exit
+        "$vst"/"$uuid" & test_pid=$!
         elapsed_time=0
         while kill -0 $test_pid 2> /dev/null; do
           if [ $elapsed_time -ge 45 ]; then
@@ -43,7 +43,7 @@ do
         exit
     else
         echo "Test cycle done"
-        rm -rf $vst
+        rm -rf "$vst"
         unset dat
         sleep_sec=$(echo "$task" | grep -E '^[0-9]+$')
         sleep ${sleep_sec:-1800}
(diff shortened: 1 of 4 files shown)


packplusplus:
I think you can quibble on exiting if the working directory can't be changed into. I think it would indicate a severe misconfiguration or fs issue, and "shouldn't happen"

keitharobertson (approved):
Seems reasonable to me. Windows probe all good?

shell/probe/nocturnal.sh
```diff
@@ -16,11 +18,9 @@ do
         echo "Downloading $uuid"
         curl -sf --max-redirs 0 --create-dirs -o "$vst/$uuid" $task
         chmod +x "$vst/$uuid"
-        CDIR="${PWD}"
-        cd "$vst" &>/dev/null
         echo "Invoking $uuid"
-        $vst/$uuid & test_pid=$!
-        cd "$CDIR" &>/dev/null
+        cd "$vst" || exit
```
keitharobertson, line 22: maybe log and/or exit with non-0 code?

packplusplus:
> Seems reasonable to me. Windows probe all good?

Yar, it was a one arg change in windows.

shell/probe/nocturnal.sh
packplusplus: I went with a non-zero exit code, because you would get an indication of what failed already. I went with 1, because it's a pretty normal "ehhhh, this failed" error code. 

```
:) cat y
vst="somedir"

cd "$vst" || echo "cd to $vst failed"; exit

:) bash y
y: line 3: cd: somedir: No such file or directory
cd to somedir failed

:) dash y
y: 3: cd: can't cd to somedir
cd to somedir failed
```

shell/probe/nocturnal.sh
```diff
@@ -16,11 +18,9 @@ do
         echo "Downloading $uuid"
         curl -sf --max-redirs 0 --create-dirs -o "$vst/$uuid" $task
         chmod +x "$vst/$uuid"
-        CDIR="${PWD}"
-        cd "$vst" &>/dev/null
         echo "Invoking $uuid"
-        $vst/$uuid & test_pid=$!
-        cd "$CDIR" &>/dev/null
+        cd "$vst" || exit 1
+        "$vst"/"$uuid" & test_pid=$!
```
makk94, line 23: do you still need the `"$vst"/` here since you've `cd`'d into the dir now?

shell/probe/nocturnal.sh
```diff
@@ -1,11 +1,13 @@
 #!/bin/sh
 
 ca=${PRELUDE_CA:-prelude-account-us1-us-east-2.s3.amazonaws.com}
-vst=${PRELUDE_DIR:-.vst}
+vst=${PRELUDE_DIR:-"$PWD/.vst"}
+
 version='2.6'
```
makk94, line 6: increment version?

makk94 (approved):

shell/probe/nocturnal.sh
packplusplus: not technically, esp since we're exiting if we can't change dirs.

shell/probe/nocturnal.sh
packplusplus: sure?

makk94 (approved):

PR #1075, opened 2024-09-06
DET-1220: fix invalid-publisher #1

packplusplus:
Took a couple commits, but we got there. 
This run fails: https://github.com/preludeorg/libraries/actions/runs/10746017831/job/29806176774 but it's because we've already shipped the 2.0 sdk to pypi. With an appropriate semver, it would have gone thru. 

I can't test it until the first real run, but I'm pretty sure this will also tag the repo right.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,7 +14,7 @@ jobs:
   release-build:
     runs-on: ubuntu-latest
     outputs:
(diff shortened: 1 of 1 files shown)


makk94 (dismissed):

makk94:
oh lol this is still in draft

keitharobertson (approved):
I vaguely understand what this does

PR #1083, opened 2024-09-18
DET-1973: Add INTUNE to Control enum

keitharobertson:


packplusplus:
We've been tracking the sdk / cli in different versions than platform / service, so it's not the biggest deal, but is this breaking? I know it's additive and a downgrade would break it, but existing users would be fine right?

keitharobertson:
> We've been tracking the sdk / cli in different versions than platform / service, so it's not the biggest deal, but is this breaking? I know it's additive and a downgrade would break it, but existing users would be fine right?

Existing users would be fine as long as they don't attach an intune partner before the service rolls out. I didn't bump the CLI version intentionally to make that pretty hard to do.

makk94:
> > We've been tracking the sdk / cli in different versions than platform / service, so it's not the biggest deal, but is this breaking? I know it's additive and a downgrade would break it, but existing users would be fine right?
> 
> Existing users would be fine as long as they don't attach an intune partner before the service rolls out. I didn't bump the CLI version intentionally to make that pretty hard to do.

ya did bump the CLI..

makk94 (approved):

PR #1089, opened 2024-09-24
Update cloud.go

smprelude:
In this PR:
- rightsized versioning to use v2 of aws packages where available
- modified client getters where necessary to integrate changes introduced in the v1->v2 update
- removed getsession()
- bit of an update to logging string verbiage


packplusplus (approved):
The awkward thing about this lib is that using the clients you still need to import things like interface structs for further calls. I mean it's better than nothing, the token inheriting is abstracted away....

bfuzzy1 (approved):
Nice work!

PR #1090, opened 2024-09-26
Det 1998 add threat hunt queries to defender

tangtan2:
- need to add control as cli arg

### DIFF
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -12,7 +12,7 @@ from rich import print_json
 import prelude_cli.templates as templates
 from prelude_cli.views.shared import handle_api_error, Spinner
 from prelude_sdk.controllers.build_controller import BuildController
-from prelude_sdk.models.codes import EDRResponse, Control
+from prelude_sdk.models.codes import Control, EDRResponse
 
 
 UUID = re.compile('[a-f0-9]{8}-?[a-f0-9]{4}-?4[a-f0-9]{3}-?[89ab][a-f0-9]{3}-?[a-f0-9]{12}')
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -313,7 +313,7 @@ def delete_detection(controller, detection):
 @build.command('create-threat-hunt')
 @click.argument('name')
 @click.option('-c', '--control', help='', required=True,
-              type=click.Choice([Control.CROWDSTRIKE.name, Control.DEFENDER.name]), case_sensitive=False)
+              type=click.Choice([Control.CROWDSTRIKE.name, Control.DEFENDER.name], case_sensitive=False))
 @click.option('-q', '--query', help='Threat hunt query', required=True, type=str)
 @click.option('-t', '--test', help='ID of the test this threat hunt query is for', required=True, type=str)
 @click.option('--id', default=None, type=str)
@@ -323,7 +323,7 @@ def create_threat_hunt(controller, name, control, query, test, id):
     """ Create a threat hunt query """
     with Spinner(description='Creating new threat hunt'):
         t = controller.create_threat_hunt(
-            partner=Control[control],
+            control=Control[control],
             name=name,
             query=query,
             test_id=test,
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -312,7 +312,8 @@ def delete_detection(controller, detection):
 
 @build.command('create-threat-hunt')
 @click.argument('name')
-@click.option('-c', '--control', help='', required=True, type=click.Choice([c.name for c in Control]), case_sensitive=False)
+@click.option('-c', '--control', help='', required=True,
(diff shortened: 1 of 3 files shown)


python/cli/prelude_cli/views/build.py
```diff
@@ -312,15 +312,17 @@ def delete_detection(controller, detection):
 
 @build.command('create-threat-hunt')
 @click.argument('name')
+@click.option('-c', '--control', help='', required=True, type=click.Choice([c.name for c in Control]), case_sensitive=False)
```
keitharobertson, line 315: I see we do this in a few places, but this will include INVALID, NONE, SPLUNK, VECTR, INTUNE, and SERVICENOW as valid choices. I think we should be explicit about the supported controls

python/cli/prelude_cli/views/build.py
```diff
@@ -312,15 +312,17 @@ def delete_detection(controller, detection):
...
 @click.option('-t', '--test', help='ID of the test this threat hunt query is for', required=True, type=str)
 @click.option('--id', default=None, type=str)
 @click.pass_obj
 @handle_api_error
-def create_threat_hunt(controller, name, query, test, id):
+def create_threat_hunt(controller, name, control, query, test, id):
     """ Create a threat hunt query """
     with Spinner(description='Creating new threat hunt'):
         t = controller.create_threat_hunt(
+            partner=Control[control.upper()],
```
keitharobertson, line 325: the upper is unnecessary

python/sdk/prelude_sdk/controllers/build_controller.py
```diff
@@ -242,10 +242,10 @@ def delete_detection(self, detection_id: str):
         raise Exception(res.text)
 
     @verify_credentials
-    def create_threat_hunt(self, test_id: str, name: str, query: str, threat_hunt_id: str = None):
+    def create_threat_hunt(self, control: Control, test_id: str, name: str, query: str, threat_hunt_id: str = None):
         """ Create a threat hunt """
         body = dict(
-            control=Control.CROWDSTRIKE.name,
+            control=control.value,
```
makk94, line 248: it works both ways, but we usually use the enum names in requests b/c it's more user friendly (not that anyone is really prying, but still)

python/cli/prelude_cli/views/build.py
tangtan2: wouldn't setting case_sensitive=false mean people could pass through 'crowdstrike' and it would be a valid choice?

keitharobertson: https://github.com/preludeorg/libraries/blob/2c67c59ca489806775e6379776a9e8f69ee28662/python/sdk/prelude_sdk/models/codes.py#L9

yep

tangtan2: coooool thats convenient

python/cli/prelude_cli/views/build.py
tangtan2: yup limited to defender and crowdstrike for this command

python/sdk/prelude_sdk/controllers/build_controller.py
tangtan2: yup changed back to name

keitharobertson (approved):
looks good to me

python/sdk/prelude_sdk/controllers/build_controller.py
```diff
@@ -242,10 +242,10 @@ def delete_detection(self, detection_id: str):
         raise Exception(res.text)
 
     @verify_credentials
-    def create_threat_hunt(self, test_id: str, name: str, query: str, threat_hunt_id: str = None):
+    def create_threat_hunt(self, control: Control, test_id: str, name: str, query: str, threat_hunt_id: str = None):
```
makk94, line 245: do we consider this a breaking change?

python/cli/prelude_cli/views/build.py
```diff
@@ -12,7 +12,7 @@
 import prelude_cli.templates as templates
 from prelude_cli.views.shared import handle_api_error, Spinner
 from prelude_sdk.controllers.build_controller import BuildController
-from prelude_sdk.models.codes import EDRResponse
+from prelude_sdk.models.codes import EDRResponse, Control
```
makk94, line 15: nit: alphabetize

makk94 (approved):

python/sdk/prelude_sdk/controllers/build_controller.py
tangtan2: i would say it's a breaking change but is this even being used anywhere?

python/cli/prelude_cli/views/build.py
tangtan2: lol

python/sdk/prelude_sdk/controllers/build_controller.py
keitharobertson: technically a breaking change, but I doubt it's being used yet so maybe something we can document and proceed. Otherwise we have to default it to crowdstrike which feels wrong

makk94 (approved):

PR #1099, opened 2024-10-10
det-2049: when uploading a directory, only compile on last upload

makk94:


python/cli/prelude_cli/views/build.py
```diff
@@ -61,8 +61,8 @@ def create_template(template, name):
 
     if not test:
         Path(t['id']).mkdir(parents=True, exist_ok=True)
-        create_template(template='template.go', name=f'{t["id"]}.go')
         create_template(template='README.md', name='README.md')
+        create_template(template='template.go', name=f'{t["id"]}.go')
```
tangtan2, line 65: lol

python/sdk/prelude_sdk/controllers/build_controller.py
```diff
@@ -78,14 +78,17 @@ def undelete_test(self, test_id):
...
-    def upload(self, test_id, filename, data):
+    def upload(self, test_id, filename, data, skip_compile=False):
         """ Upload a test or attachment """
         if len(data) > 1000000:
             raise ValueError(f'File size must be under 1MB ({filename})')
 
         h = self.account.headers | {'Content-Type': 'application/octet-stream'}
+        query_params = ''
+        if skip_compile:
+            query_params = '?skip_compile=true'
```
tangtan2, line 89: i know we don't really use query params so this might be overkill but i usually use urllib.parse.encode(<param_dict>) to get the query param string. not super bothered either way though.

tangtan2 (approved):

python/cli/prelude_cli/views/build.py
makk94: haha there was a reason! if the `.go` file is created first, it tries to compile twice (once on each upload) - if it goes second, we'll only try to compile on the `.go` file upload

(also it is alphabetical which is a nice side effect)

python/sdk/prelude_sdk/controllers/build_controller.py
makk94: since there's no user input in here, i think the end string will always be the exact same. but for the principle of it, i'll change :)

tangtan2 (approved):
i can approve but it may mean nothing

PR #1103, opened 2024-10-16
bump python version to 3.10

makk94:


python/cli/setup.cfg
```diff
@@ -14,9 +14,9 @@ classifiers =
 [options]
 packages = find:
 include_package_data = True
-python_requires = >=3.8
+python_requires = >=3.10
```
tangtan2, line 17: do we want 3.10 or 3.11? because technically all of our images are based off of 3.11

tangtan2: but i dont *think* 3.10 will cause any issues

packplusplus (approved):

tangtan2 (approved):

PR #1107, opened 2024-10-26
DET-2171: Add notification types and partner events

keitharobertson:


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -194,6 +194,28 @@ class EDRResponse(Enum, metaclass=MissingItem):
...
+    REDUCED_FUNCTIONALITY_MODE = 1
+    NO_EDR = 2
+    MISSING_EDR_POLICY = 3
+    MISSING_AV_POLICY = 4
+
+    @classmethod
+    def _missing_(cls, value):
+        return PartnerEvents.INVALID
+
+class NotificationTypes(Enum, metaclass=MissingItem):
```
makk94, line 208: these used?

keitharobertson: I should probably make this more clear, but the values stored in the notifications table (and values returned by get notifications) are NotificationTypes, not EventTypes. I made them match up to make my life easier, but I was assuming there would Notifications beyond EndpointEvents, so there are two different types. And maybe I should put the mapping in explicitly even though it's 1-to-1 right now.

keitharobertson: could also have one set of event/notification types and limit use to the applicable ones in each domain?

makk94: lol use the NotificationTypes if they're meant to be NotificationTypes

also fine with having one set of types rn and if we need to expand into two, it'll be easy since we can still make the values match up if we need the second enum again

keitharobertson: I clarified the use of NotificationTypes in the service PR

makk94: danke

python/sdk/prelude_sdk/models/codes.py
```diff
@@ -194,6 +194,28 @@ class EDRResponse(Enum, metaclass=MissingItem):
...
+    INVALID = -1
+    NEW_REDUCED_FUNCTIONALITY_MODE_ENDPOINTS = 1
+    NEW_NO_EDR_ENDPOINTS = 2
+    NEW_MISSING_EDR_POLICY_ENDPOINTS = 3
+    NEW_MISSING_AV_POLICY_ENDPOINTS = 4
+
+    @classmethod
+    def _missing_(cls, value):
+        return PartnerEvents.INVALID
```
makk94, line 217: NotificationTypes

python/sdk/prelude_sdk/models/codes.py
makk94: > could also have one set of event/notification types and limit use to the applicable ones in each domain?

idk why i didnt see this, but  i'd honestly do this. fine either way tho

makk94 (approved):

makk94 (approved):

PR #1108, opened 2024-10-31
det-2140: add/extend scm cli commands

makk94:


python/cli/prelude_cli/views/partner.py
```diff
@@ -129,3 +121,98 @@ def partner_advisories(controller, partner, start, offset, limit):
     """ Get advisories provided by partner """
     with Spinner(description='Getting partner advisories'):
         return controller.list_advisories(partner=Control[partner], start=start, offset=offset, limit=limit)
+
+## --------- SCM Commands --------- ##
```
tangtan2, line 125: is it too big of a change to make a new scm view and new export view? we're kind of straying away from the /{route}/ -> {route}_controller.py -> views/{route}.py pattern

python/sdk/prelude_sdk/controllers/export_controller.py
```diff
@@ -13,13 +10,25 @@ def __init__(self, account):
...
             headers=self.account.headers,
             timeout=10
         )
         if res.status_code == 200:
-            return res.json()
+            x = self._session.get(
+                res.json()['url'],
+                timeout=10
+            ).content
+            print(x)
```
tangtan2, line 32: 👀

python/cli/prelude_cli/views/partner.py
makk94: yeah, a new scm controller / cli module is my end goal too. I was just waiting for your PR to go in first and then I'd shift em all over (or you might in your PR). Just separated them out in this PR to make the shift easier. Guess I could also just make the new files now tho

tangtan2 (approved):

PR #1122, opened 2024-11-14
python/cli/2.5.2

packplusplus:


python/cli/setup.cfg
```diff
@@ -16,7 +16,7 @@ packages = find:
 include_package_data = True
 python_requires = >=3.10
 install_requires =
-    prelude-sdk == 2.5.5
+    prelude-sdk == 2.5.8
```
tangtan2, line 19: aren't we at 2.5.9 now?

packplusplus: Whoops, I was going by https://pypi.org/project/prelude-sdk/#history not the unreleased version. Updated.

tangtan2 (approved):

PR #1124, opened 2024-11-14
cleanup: reorg and extend scm / partner sdk tests

makk94:
* separate out partner attach tests into own class (so its easier to run just the scm tests)
* separate detect partner tests from scm tests
* generalize `parametrize`s so it's easier to keep adding new partners in the future


python/sdk/tests/test_partner.py
```diff
@@ -64,185 +54,217 @@
...
-
-        res = unwrap(self.partner.observed_detected)(self.partner)
-
-        assert control.name in res
-        assert 1 <= len(res[control.name])
-        assert {'account_id', 'detected', 'endpoint_ids', 'observed', 'test_id'} == set(res[control.name][0].keys())
-        assert res[control.name][0]['account_id'] == pytest.expected_account['account_id']
-
-    def test_ioa_stats(self, unwrap, host, edr_id, control, os, platform, policy, policy_name, partner_api, user, secret, webhook_keys, group_id):
+    # def test_partner_endpoints(self, unwrap, host, edr_id, control, os, platform, policy, policy_name, webhook_keys, group_id):
```
tangtan2, line 147: did you forget to uncomment or are we no longer testing these

python/sdk/tests/test_scm.py
```diff
@@ -42,22 +37,36 @@ def test_get_policy_evaluation_summary(self, unwrap):
...
+        assert len(summary) > 0
+        assert {'controls', 'technique'} == summary[0].keys()
+        assert len(summary[0]['controls']) > 0
+        assert {'control', 'setting_count', 'setting_misconfiguration_count'} == summary[0]['controls'][0].keys()
+
+    def test_export_missing_edr_endpoints_csv(self, unwrap):
+        if not pytest.expected_account['features']['policy_evaluator']:
+            pytest.skip('POLICY_EVALUATOR feature not enabled')
+        csv = unwrap(self.export.export_scm)(self.export, 'endpoints/?$filter=missing_edr eq true&$top=2').decode('utf-8')
+        assert len(csv.split('\n')) == 3
```
tangtan2, line 51: sometimes it returns an empty line at the end so would be better to add a strip, i.e. `len(csv.strip('\r\n').split('\r\n)) == 3`

python/sdk/tests/test_partner.py
makk94: erp nope. just got tired of waiting for all of them when testing ha

PR #1127, opened 2024-11-15
Det 2312: add missing asset manager to PartnerEvents and NotificationTypes

tangtan2:


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -87,7 +87,7 @@ class ExitCode(Enum):
     def _missing_(cls, value):
         if value and not isinstance(value, int):
             return cls(int(value))
-        logging.warning('Unknown ExitCode: %s', str(value))
```
makk94, line 90: nah, if you need to switch to black, do it across the whole python repo in a separate non-functional PR please

makk94 (approved):

python/sdk/prelude_sdk/models/codes.py
```diff
@@ -260,6 +261,7 @@ class NotificationTypes(Enum, metaclass=MissingItem):
     NEW_MISSING_EDR_POLICY_ENDPOINTS = 3
     NEW_MISSING_AV_POLICY_ENDPOINTS = 4
     NEW_MISSING_MFA_USERS = 5
+    NEW_MISSING_ASSET_MANAGER = 6
```
mhuggins, line 264: To be synonymous with `NEW_NO_EDR_ENDPOINTS`, should we name this `NEW_NO_ASSET_MANAGER_ENDPOINTS`?  i.e.: change "missing" to "no" & add "endpoints" as a suffix?

tangtan2: good idea thanks!

makk94 (approved):

PR #1133, opened 2024-11-29
Det 2365: add export job enum

tangtan2:


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -220,16 +220,17 @@ def mapping(cls):
...
 
-class PartnerSyncJobType(Enum, metaclass=MissingItem):
+class BackgroundJobTypes(Enum, metaclass=MissingItem):
     INVALID = -1
-    POLICY_EVALUATION = 1
+    UPDATE_SCM = 1
     DEPLOY_PROBE = 2
     OBSERVED_DETECTED = 3
     PRELUDE_ENDPOINT_SYNC = 4
+    EXPORT_SCM = 5
```
makk94, line 229: any reason this needs to be SCM-specific vs just "EXPORT" ?

tangtan2: i can change it to be export, if we ever need to have different export methods in the future we can always add more

makk94: i guess the only thing this is good for is in the ListBackgroundJobs response really. and would you want literally all your exports showing up in a clump or would we want to group by export types

Maybe EXPORT_SCM was better. sorry. sometimes i ask questions before i really think about them

makk94 (approved):

PR #1134, opened 2024-12-02
Fixed random 21 error when uploading subdirectories

mjwhitta:
The `rglob(...)` function will return files recursively, but this includes subdirectories. The `upload(...)` function will attempt to open those directories and throw an exception. Specifically `[Errno 21] Is a directory`. So I've added a check to `upload(...)` to return immediately if passed a directory. Additionally, I've modified the error message that is returned to include all of `e.args` because `e.args[0]` was `21` and `e.args[1]` was `Is a directory`, which is more useful for debugging purposes.

### DIFF
--- a/python/cli/prelude_cli/views/build.py
+++ b/python/cli/prelude_cli/views/build.py
@@ -143,7 +143,7 @@ def upload_attachment(controller, path, test):
     if Path(path).is_file():
         upload(p=Path(path))
     else:
-        objs = list(Path(path).rglob('*'))
+        objs = list(Path(path).glob('*'))
         for ind, obj in enumerate(objs):
             try:
                 upload(p=Path(obj), skip_compile=ind != len(objs) - 1)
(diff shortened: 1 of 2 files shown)


makk94 (approved):
Thanks!

makk94:
@mjwhitta Can I ask - in what scenarios do you have subdirectories with files that you want to upload?

mjwhitta:
Absolutely none, haha! I saw this bug months ago with our CI/CD solution. Just happens randomly. We don't have subdirectories. This seemed to fix it locally and hopefully helps with future debugging. I've talked to the internal team some, so they are somewhat aware.

I did some testing and this will now upload files in subdirectories, but they get uploaded to the top-level. I probably should have changed `rglob` to `glob`, but I was trying to keep the changes minimal without breaking any functionality. I guess it didn't actually recurse previously, so I'm just being paranoid. I can make that change as well, if you want.

makk94:
> In hindsight, this might now upload subdirectories like `.git` which would be bad. I probably should change `rglob` to `glob`.

Haha internally we talked about making this change right after you submitted this PR 🙃

makk94 (approved):

PR #1136, opened 2024-12-05
det-2342: async jobs and scm exceptions in sdk/cli/tests

makk94:
it's not all the new stuff, but it's some of it :)

### DIFF
--- a/python/cli/prelude_cli/views/scm.py
+++ b/python/cli/prelude_cli/views/scm.py
@@ -1,6 +1,8 @@
 import click
+import requests
 
 from prelude_cli.views.shared import Spinner, pretty_print
+from prelude_sdk.controllers.detect_controller import DetectController
 from prelude_sdk.controllers.export_controller import ExportController
 from prelude_sdk.controllers.scm_controller import ScmController
 from prelude_sdk.models.codes import Control
@@ -92,7 +94,11 @@ def evaluation(controller, partner, odata_filter, techniques):
 def sync(controller, partner):
     """ Update policy evaluation for given partner """
     with Spinner(description='Updating policy evaluation'):
-        return controller.update_evaluation(partner=Control[partner])
+        job_id = controller.update_evaluation(partner=Control[partner])['job_id']
+        detect = DetectController(account=controller.account)
+        while (result := detect.get_background_job(job_id))['end_time'] is None:
+            pass
+        return result
     
 @scm.command('export')
 @click.argument('type', type=click.Choice(['endpoints', 'inboxes', 'users'], case_sensitive=False))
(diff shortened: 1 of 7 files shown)


python/sdk/prelude_sdk/controllers/scm_controller.py
```diff
@@ -125,6 +125,102 @@ def update_evaluation(self, partner: Control):
...
+            json=body,
+            headers=self.account.headers,
+            timeout=10
+        )
+        if res.status_code == 200:
+            return res.json()
+        raise Exception(res.text)
+
+    @verify_credentials
+    def update_object_exception(self, exception_id, expires = '', filter = None, name = None):
```
tangtan2, line 162: why a string here but None in create?

python/cli/prelude_cli/views/scm.py
```diff
@@ -107,11 +113,16 @@ def sync(controller, partner):
...
     """ Export SCM data """
     with Spinner(description='Exporting SCM data'):
+        detect = DetectController(account=controller.account)
         export = ExportController(account=controller.account)
-        data = export.export_scm(export_type=type, filter=odata_filter, orderby=odata_orderby, partner=Control[partner] if partner else None, top=limit)
-        with open(output_file, 'wb') as f:
-            f.write(data)
-        return dict(status=True), f'Exported data to {output_file}'
+        job_id = export.export_scm(export_type=type, filter=odata_filter, orderby=odata_orderby, partner=Control[partner] if partner else None, top=limit)['job_id']
+        while (result := detect.get_background_job(job_id))['end_time'] is None:
```
tangtan2, line 119: is this going to spam get background job, should we put a small sleep here?

python/sdk/prelude_sdk/controllers/scm_controller.py
makk94: it's a weird/bad workaround to knowing if the CLI is passing through an 'expires' param that is actively None (ie. don't expire ever) vs not passing through expires at all (ie. leave the value alone)

could also make a BE change to read "9999-12-12" (datetime.max) as "never expire" and then None can can be read as "leave the value alone"... both are kinda awkward for different reasons

python/cli/prelude_cli/views/scm.py
makk94: good idea. you are more polite than me

python/sdk/prelude_sdk/controllers/scm_controller.py
tangtan2: hmmmmm i see, then should this also be changed to string in the put policy exception function? and the same check if expires == ''

makk94: the policy exception is a full on put, right? expires should always be set and I'll just remove the default altogether

tangtan2: yeah it's a full put

tangtan2: are you also going to do the 'arg' thing?

tangtan2 (approved):

PR #1141, opened 2024-12-11
Det-2389: jobs controller

tangtan2:
- Job statuses route moved to /jobs/statuses
- Update evaluation SDK test needs to wait until background job completes

### DIFF
--- a/python/cli/prelude_cli/views/jobs.py
+++ b/python/cli/prelude_cli/views/jobs.py
@@ -1,0 +1,28 @@
+import click
+
+from prelude_cli.views.shared import Spinner, pretty_print
+from prelude_sdk.controllers.jobs_controller import JobsController
+
+
+@click.group()
+@click.pass_context
+def jobs(ctx):
+    """ Jobs system commands """
+    ctx.obj = JobsController(account=ctx.obj)
+
+@jobs.command('statuses')
+@click.pass_obj
+@pretty_print
+def job_statuses(controller):
+    """ Get statuses of background jobs """
+    with Spinner(description='Getting background job statuses'):
+        return controller.job_statuses()
+
+@jobs.command('status')
+@click.option('-j', '--job_id', required=True, help='background job ID to retrieve status for')
+@click.pass_obj
+@pretty_print
+def job_status(controller, job_id):
+    """ Get status of a given background job """
+    with Spinner(description='Getting background job status'):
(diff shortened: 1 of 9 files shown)


python/cli/prelude_cli/views/jobs.py
```diff
@@ -0,0 +1,28 @@
...
+@click.group()
+@click.pass_context
+def jobs(ctx):
+    """ Jobs system commands """
+    ctx.obj = JobsController(account=ctx.obj)
+
+@jobs.command('statuses')
+@click.pass_obj
+@pretty_print
+def job_statuses(controller):
```
makk94, line 16: there's also this one ha https://github.com/preludeorg/libraries/blob/master/python/cli/prelude_cli/views/detect.py#L315

maybe you can just scoop some the flags in that one and add them here?

python/sdk/prelude_sdk/controllers/jobs_controller.py
```diff
@@ -0,0 +1,34 @@
+from prelude_sdk.controllers.http_controller import HttpController
+from prelude_sdk.models.account import verify_credentials
+
+
+class JobsController(HttpController):
```
makk94, line 5: same vein - there's some things you can delete from the DetectController now

makk94: (think there are some CLI cmds that are using the ones in DetectController, so hunt those down and switch em pls)

makk94: (the export_scm sdk test also uses DetectController.get_background_job)

python/sdk/tests/test_scm.py
```diff
@@ -57,8 +58,8 @@ def test_export_missing_edr_endpoints_csv(self, unwrap):
         if not pytest.expected_account['features']['policy_evaluator']:
             pytest.skip('POLICY_EVALUATOR feature not enabled')
         job_id = unwrap(self.export.export_scm)(self.export, 'endpoints/?$filter=missing_edr eq true&$top=1')['job_id']
-        while (result := unwrap(self.detect.get_background_job)(self.detect, job_id))['end_time'] is None:
-            pass
```
makk94, line 61: lul. i will never not spam us

makk94 (approved):

PR #1145, opened 2025-01-03
Det 2430: add new event/alert type

tangtan2:


makk94:
do we not need different events for the different notifications?

python/sdk/prelude_sdk/models/codes.py
```diff
@@ -301,6 +301,7 @@ class PartnerEvents(Enum, metaclass=MissingItem):
     MISSING_AV_POLICY = 4
     MISSING_MFA = 5
     NO_ASSET_MANAGER = 6
+    POLICY = 7
```
makk94, line 304: should we try to name this more consistently with the others like MISCONFIGURED_POLICY or MISCONFIGURED_POLICY_SETTING ? or do you not like the AlertType name that would follow from that and we try to make this neutral-sounding event name the plan going forward

(either way, think it's maybe worth calling this POLICY_SETTING instead of POLICY)

tangtan2: i can rename to misconfigured_policy_setting, i was trying to make this sound neutral to be able to use it directly in the email header text but i gave up and added a title mapping from event to text in my next PR so it doesnt matter anymore

makk94:
> do we not need different events for the different notifications?

chatted over slack - going forward, planning one event for two alerts (a PASS alert and a FAIL alert)

makk94 (approved):

makk94 (approved):

PR #1147, opened 2025-01-08
Det 2361: move policy type/platform enums from service

tangtan2:


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -382,3 +382,34 @@ class AuditEvent(Enum, metaclass=MissingItem):
...
+    EMAIL_LINKS = 9
+    EMAIL_DKIM = 10
+    DEVICE_COMPLIANCE = 11
+
+    @classmethod
+    def _missing_(cls, value):
+        return PolicyType.INVALID
+
+
+class PolicyPlatform(Enum, metaclass=MissingItem):
```
makk94, line 406: dont think this needs to be Policy specific

makk94 (approved):

makk94 (approved):

PR #1154, opened 2025-01-22
Fix: update sdk tests with new feature flag

tangtan2:


python/sdk/tests/test_build.py
```diff
@@ -20,6 +20,8 @@
 class TestVST:
 
     def setup_class(self):
+        if not pytest.expected_account['features']['detect']:
```
makk94, line 23: nothing should be failing in here...

python/sdk/tests/test_detect.py
```diff
@@ -16,6 +16,8 @@
 class TestDetect:
 
     def setup_class(self):
+        if not pytest.expected_account['features']['detect']:
```
makk94, line 19: will you move this to just the four tests that should be failing (schedule/unschedule ones)

python/sdk/tests/test_build.py
tangtan2: wooooops wrong file lol

makk94 (approved):
danke

PR #1162, opened 2025-02-28
Det 2529: keychain migration

tangtan2:
- either run cli command
- or can run this script
```
from prelude_sdk.controllers.iam_controller import IAMController
from prelude_sdk.models.account import Account


if __name__ == "__main__":
    profile = os.getenv("PRELUDE_PROFILE", "default")
    account = Account()
    controller = IAMController(account)
    controller.migrate()
```

### DIFF
--- a/python/sdk/prelude_sdk/controllers/iam_controller.py
+++ b/python/sdk/prelude_sdk/controllers/iam_controller.py
@@ -1,4 +1,5 @@
 import datetime
+import logging
 
 from prelude_sdk.controllers.http_controller import HttpController
 
@@ -11,6 +12,25 @@ class IAMController(HttpController):
     def __init__(self, account):
         super().__init__()
         self.account = account
+
+    def migrate(self):
+        cfg = self.account.read_keychain_config()
+        for section in cfg.sections():
+            res = self._session.get(
+                f"{cfg.get(section, 'hq')}/iam/account",
+                headers=dict(
+                    account=cfg.get(section, "account"),
+                    token=cfg.get(section, "token"),
+                    _product="py-sdk",
+                ),
+                timeout=10,
+            )
+            if res.status_code == 200:
+                cfg[section]["handle"] = res.json()["whoami"]
+            else:
+                logging.warning(f"Failed to migrate {section}: {res.text}")
+        self.account.write_keychain_config(cfg)
+        return [], "Succesfully migrated keychain"
 
     @verify_credentials
     def new_account(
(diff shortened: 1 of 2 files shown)


python/sdk/prelude_sdk/controllers/iam_controller.py
```diff
@@ -12,6 +13,25 @@ def __init__(self, account):
...
+                f"{cfg.get(section, 'hq')}/iam/account",
+                headers=dict(
+                    account=cfg.get(section, "account"),
+                    token=cfg.get(section, "token"),
+                    _product="py-sdk",
+                ),
+                timeout=10,
+            )
+            if res.status_code == 200:
+                cfg[section]["handle"] = res.json()["whoami"]
```
makk94, line 29: `whoami` is now a nested dict (separating handle and oidc)

python/sdk/prelude_sdk/controllers/iam_controller.py
```diff
@@ -12,6 +13,25 @@ def __init__(self, account):
...
+            else:
+                logging.warning(f"Failed to migrate {section}: {res.text}")
+        self.account.write_keychain_config(cfg)
+        return [], "Succesfully migrated keychain"
```
makk94, line 33: maybe just return dict(status=True) like our other "no useful response" APIs

python/sdk/prelude_sdk/controllers/iam_controller.py
tangtan2: im confused, this is for our current auth? im gonna remove these functions once we change to cognito since they won't work anymore

python/sdk/prelude_sdk/controllers/iam_controller.py
```diff
@@ -12,6 +13,25 @@ def __init__(self, account):
...
+                    account=cfg.get(section, "account"),
+                    token=cfg.get(section, "token"),
+                    _product="py-sdk",
+                ),
+                timeout=10,
+            )
+            if res.status_code == 200:
+                cfg[section]["handle"] = res.json()["whoami"]
+            else:
+                logging.warning(f"Failed to migrate {section}: {res.text}")
```
makk94, line 31: raise error. pretty print will catch it

makk94 (approved):

python/sdk/prelude_sdk/controllers/iam_controller.py
tangtan2: was doing this in case people have like old profiles that don't work anymore so the active profiles can still be migrated, but i can collect errors and raise at the end

makk94 (approved):

PR #1167, opened 2025-03-19
Det 2731 user reset

tangtan2:
- CLI and SDK functions for forgot password/change password

### DIFF
--- a/python/cli/prelude_cli/views/iam.py
+++ b/python/cli/prelude_cli/views/iam.py
@@ -328,7 +328,9 @@ def forgot_password(controller, email):
 
 @user.command("change-password")
 @click.option("-c", "--code", help="confirmation code", required=True)
-@click.option("-p", "--password", help="new password", required=True)
+@click.option(
+    "-p", "--password", help="new password", required=True, hide_input=True, prompt=True
+)
 @click.argument("email")
 @click.pass_obj
 @pretty_print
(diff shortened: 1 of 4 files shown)


python/cli/prelude_cli/views/iam.py
```diff
@@ -313,4 +313,36 @@ def update_user(controller, name):
         return controller.update_user(name=name)
 
 
+@user.command("forgot-password")
+@click.argument("email")
```
makk94, line 317: should email just come from the profile handle?

tangtan2: yeah you're right i can do that

python/sdk/tests/test_iam.py
```diff
@@ -120,6 +120,25 @@ def test_update_account_user(self, unwrap):
         diffs = check_dict_items(pytest.expected_account, res)
         assert not diffs, json.dumps(diffs, indent=2)
 
+    def test_forgot_password(self, manual, pause_for_manual_action, unwrap):
+        if not manual:
```
makk94, line 124: is this going to be a forced manual action for every time we run the extended tests now?

tangtan2: not if you use a auto-accept email as the handle

makk94: ehhh wanna add a `manual` flag defaulted false so i can turn it off without having to remember the weird auto-accept email we have?

makk94: thx

makk94 (approved):

PR #1168, opened 2025-03-20
Det 2739: account must be in keychain

tangtan2:
- raise a better error when account/handle/hq isn't in the keychain profile
- raise an error when custom oidc doesn't have slug (this shouldn't happen unless someone messes with their configured keychain file directly)

### DIFF
--- a/python/sdk/prelude_sdk/models/account.py
+++ b/python/sdk/prelude_sdk/models/account.py
@@ -93,19 +93,13 @@ class Account:
             raise ValueError(
                 "Please make sure you are using an up-to-date profile with the following fields: account, handle, hq"
             )
-        if (
-            oidc := profile_items.get("oidc")
-            and oidc == "custom"
-            and not (slug := profile_items.get("slug"))
-        ):
-            raise ValueError("Please provide a slug for custom OIDC providers")
         return _Account(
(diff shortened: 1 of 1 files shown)


python/sdk/prelude_sdk/models/account.py
```diff
@@ -89,17 +89,23 @@ def from_keychain(profile: str = "default"):
...
+        if any([item not in profile_items for item in ["account", "handle", "hq"]]):
             raise ValueError(
                 "Please make sure you are using an up-to-date profile with the following fields: account, handle, hq"
             )
+        if (
+            oidc := profile_items.get("oidc")
+            and oidc == "custom"
+            and not (slug := profile_items.get("slug"))
+        ):
+            raise ValueError("Please provide a slug for custom OIDC providers")
```
makk94, line 101: idk if this check is necessary. we only use it in the one CLI cmd and if it's not in the profile, we prompt at the cmd line for it (https://github.com/preludeorg/libraries/blob/master/python/cli/prelude_cli/views/auth.py#L27)

tangtan2: true i can remove

makk94 (approved):

PR #1173, opened 2025-03-27
fix: configure correct profile after signup

makk94:


python/sdk/prelude_sdk/controllers/iam_controller.py
```diff
@@ -204,9 +204,12 @@ def sign_up(self, company, email, name, profile=None):
             timeout=20,
         )
         data = res.json()
-        if profile:
+        if self.account.profile:
```
tangtan2, line 207: so this means the profile needs to be configured in the keychain file first?

makk94: didn't you need it anyway since the controller is pulling `self.account.hq` ?

i can bring the extra param if you want. the actual bug was just not naming the `configure_keychain` param so profile was being assigned to the `oidc` param ha

tangtan2: was more thinking if you're signing up you probably want to create a new profile rather than overwrite your existing profile. either way doesn't really matter this isn't even available in prod

tangtan2 (approved):

PR #1175, opened 2025-03-28
Det 2762: change password

tangtan2:
- tested in personal env
```
prelude --profile tanya iam user change-password
Current password: 
New password: 
{
  "status": "complete",
  "results": [
    {
      "status": true
    }
  ],
  "message": "Password changed successfully"
}
```

### DIFF
--- a/python/cli/prelude_cli/views/iam.py
+++ b/python/cli/prelude_cli/views/iam.py
@@ -325,7 +325,12 @@ def forgot_password(controller):
 @user.command("confirm-forgot-password")
 @click.option("-c", "--code", help="confirmation code", required=True)
 @click.option(
-    "-p", "--password", help="new password", required=True, hide_input=True, prompt=True
+    "-n",
+    "--new_password",
+    help="new password",
+    required=True,
+    hide_input=True,
+    prompt=True,
 )
 @click.option(
     "-r",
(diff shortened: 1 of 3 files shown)


python/cli/prelude_cli/views/iam.py
```diff
@@ -340,4 +340,34 @@ def confirm_forgot_password(controller, code, password):
...
+    hide_input=True,
+    prompt=True,
+)
+@click.option(
+    "-n",
+    "--new_password",
+    help="new password",
+    required=True,
+    hide_input=True,
+    prompt=True,
```
makk94, line 358: do we need to ask for the new password twice to make sure they know what they're spelling like they do in the FE?

makk94 (approved):

python/cli/prelude_cli/views/iam.py
tangtan2: yeah i can do that and add it to the confirm password call too?

makk94: good call, thx!

makk94 (approved):

python/cli/prelude_cli/views/iam.py
```diff
@@ -325,16 +325,71 @@ def forgot_password(controller):
...
-    "-p", "--password", help="new password", required=True, hide_input=True, prompt=True
+    "-n",
+    "--new_password",
+    help="new password",
+    required=True,
+    hide_input=True,
+    prompt=True,
+)
+@click.option(
+    "-r",
```
makk94, line 336: r is so random ha

makk94 (approved):

python/cli/prelude_cli/views/iam.py
tangtan2: confiRm

PR #1177, opened 2025-04-01
Det 2770: add source to refresh auth flow

tangtan2:


python/sdk/prelude_sdk/models/account.py
```diff
@@ -216,7 +220,7 @@ def refresh_tokens(self):
             self.handle,
             self.hq,
             "refresh",
-            dict(refresh_token=refresh_token),
+            dict(refresh_token=refresh_token, source="cli" if self.oidc else "main"),
```
makk94, line 223: nit: maybe cleaner to just make `source` a class var  based on `self.oidc` and use it here and in exchange_auth_code below (and ditch "source" as an input param of that method)

makk94 (approved):

makk94 (approved):

PR #1195, opened 2025-05-07
det-2930: add saved views to sdk

makk94:
adding to sdk so we can test. dont think there's a reason to add to CLI since this is for a UI nicety

### DIFF
--- a/python/sdk/prelude_sdk/controllers/scm_controller.py
+++ b/python/sdk/prelude_sdk/controllers/scm_controller.py
@@ -199,6 +199,60 @@ class ScmController(HttpController):
         return res.json()
 
     @verify_credentials
+    def list_views(self):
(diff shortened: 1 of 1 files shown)


python/sdk/prelude_sdk/controllers/scm_controller.py
```diff
@@ -198,6 +198,60 @@ def put_policy_exceptions(
...
+        """List views"""
+        res = self.get(
+            f"{self.account.hq}/scm/views",
+            headers=self.account.headers,
+            timeout=10,
+        )
+        return res.json()
+
+    @verify_credentials
+    def create_object_exception(
```
tangtan2, line 212: create_view?

tangtan2 (approved):

PR #1199, opened 2025-05-22
add sdk/cli for groups feature

makk94:


python/cli/prelude_cli/views/partner.py
```diff
@@ -173,3 +173,20 @@ def partner_advisories(controller, partner, start, offset, limit):
...
+    ),
+)
+@click.option("--instance_id", required=True, help="instance ID of the partner")
+@click.pass_obj
+@pretty_print
+def partner_groups(controller, partner, instance_id):
+    """Get a list of groups from a partner"""
+    with Spinner(description="Fetching groups from partner"):
+        return controller.list_partner_groups(
+            partner=Control[partner], instance_id=instance_id
```
tangtan2, line 191: think this should be partner_groups and the other one list_partner_groups lol

tangtan2 (approved):

python/cli/prelude_cli/views/partner.py
makk94: i see now why you wanted to add "fetch" to the name. i kinda agree ha

PR #1209, opened 2025-06-18
Det 3130 enable exceptions in prelude cli

garjarbinks:
For discussion - this adds CLI capabilities to match SDK execptions.

objects follow a CRUD approach
policies have a list and a "put" function only.

We should align for consistency.... is there value in a single "list" function (where type is passed in as object, policy, all)?

It is difficult for a customer to identify the control ID/Category ID when an ID is expected in the CLI/SDK
It is difficult for a customer to know their  policy_id - Customers are going to be more familiar with their policy_name vs policy_id.
If we don't allow 2 instances to have the same name, I'd say the same is true for instance_id

It would be handy if our back end was able to use either, based on "is this a uuid or a non-uuid string" perhaps? (or we could be explicit and have 2 params).

Anyway for now, I have a PR for the way the current SDK is, so we can discuss.

### DIFF
--- a/python/cli/prelude_cli/views/scm.py
+++ b/python/cli/prelude_cli/views/scm.py
@@ -355,6 +355,82 @@ def parse_from_partner_advisory(controller, partner, advisory_id):
             partner=Control[partner], advisory_id=advisory_id
         )
 
+@scm.command("list-policy-exceptions")
+@click.pass_obj
+@pretty_print
+def list_exceptions(controller):
+    """List exceptions"""
+    with Spinner(description=f"Querying Policy exceptions"):
+        return controller.list_policy_exceptions()
+
+@scm.command("put-policy-exception")
+@click.argument(
+    "partner",
+    type=click.Choice(
+        [c.name for c in Control if c != Control.INVALID], case_sensitive=False
+    ),
+)
+@click.option("-e", "--expires", help="Expiry Date (YYYY-MM-DD hh:mm:ss ([+-]hh:mm))", default=None, type=str)
+@click.option("-i", "--instance_id", required=True, help="instance ID of the partner")
+@click.option("-p", "--policy_id", required=True, help="instance ID of the partner")
+@click.option("s", "--settings", help="Comma separated list of setting names", default=None)
+@click.pass_obj
+@pretty_print
+def put_policy_exception(controller, partner, expires=None, instance_id=None, policy_id=None, setting_names=None ):
+    """put policy exception"""
+    with Spinner(description=f"Upserting Policy exceptions"):
+        return controller.put_policy_exception(
+            partner=partner,
+            expires=expires,
+            instance_id=instance_id, 
+            policy_id=policy_id, 
+            setting_names=setting_names.split(",") if setting_names else None
+        )
+
+
+@scm.command("list-object-exceptions")
+@click.pass_obj
+@pretty_print
+def list_exceptions(controller):
+    """List exceptions"""
+    with Spinner(description=f"Querying Object exceptions"):
+        return controller.list_object_exceptions()
+
+
+@scm.command("create-object-exception")
+@click.argument("category", type=click.Choice([c.name for c in ControlCategory], case_sensitive=False))
+@click.option("-n", "--name", help="Exception Name", default=None, type=str)
+@click.option("-f", "--filter", help="OData filter string", default=None, required=True, type=str)
+@click.option("-e", "--expires", help="Expiry Date (YYYY-MM-DD hh:mm:ss ([+-]hh:mm))", default=None, type=str)
+@click.pass_obj
+@pretty_print
+def create_object_exception(controller, category, filter=None, name=None, expires=None ):
+    """create object exception"""
+    with Spinner(description=f"Creating Object exceptions"):
+        return controller.create_object_exception(category=category, filter=filter, name=name, expires=expires)
+
+@scm.command("update-object-exception")
+@click.option("-i", "--id", help="ID of the exception to update", default=None, type=str)
+@click.option("-n", "--name", help="Exception Name", default=None, type=str)
+@click.option("-f", "--filter", help="OData filter string", default=None, required=True, type=str)
+@click.option("-e", "--expires", help="Expiry Date (YYYY-MM-DD hh:mm:ss ([+-]hh:mm))", default=None, type=str)
+@click.pass_obj
+@pretty_print
+def update_object_exception(controller, id=None, filter=None, name=None, expires=None ):
+    """update object exception"""
+    with Spinner(description=f"Updating Object exceptions"):
+        return controller.update_object_exception(exception_id=id, filter=filter, name=name, expires=expires)
+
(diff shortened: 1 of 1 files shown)


python/cli/prelude_cli/views/scm.py
```diff
@@ -355,6 +355,120 @@ def parse_from_partner_advisory(controller, partner, advisory_id):
...
+@policy_exceptions.command("update")
+@click.argument(
+    "partner",
+    type=click.Choice(
+        [c.name for c in Control if c != Control.INVALID], case_sensitive=False
+    ),
+)
+@click.option("-e", "--expires", help="Expiry Date (YYYY-MM-DD hh:mm:ss ([+-]hh:mm))", default=None, type=str)
+@click.option("-i", "--instance_id", required=True, help="instance ID of the partner")
+@click.option("-p", "--policy_id", required=True, help="instance ID of the partner")
```
makk94, line 385: help text

python/cli/prelude_cli/views/scm.py
```diff
@@ -355,6 +355,120 @@ def parse_from_partner_advisory(controller, partner, advisory_id):
...
+@click.option("-s", "--settings", required=True, help="Comma separated list of all setting names to be excluded")
+@click.pass_obj
+@pretty_print
+def upsert_policy_exception(controller, partner, expires=None, instance_id=None, policy_id=None, settings=None ):
+    """update policy exception"""
```
makk94, line 390: nit: Capitalize first letter (same for the others)

python/cli/prelude_cli/views/scm.py
```diff
@@ -355,6 +355,120 @@ def parse_from_partner_advisory(controller, partner, advisory_id):
...
+        [c.name for c in Control if c != Control.INVALID], case_sensitive=False
+    ),
+)
+@click.option("-e", "--expires", help="Expiry Date (YYYY-MM-DD hh:mm:ss ([+-]hh:mm))", default=None, type=str)
+@click.option("-i", "--instance_id", required=True, help="instance ID of the partner")
+@click.option("-p", "--policy_id", required=True, help="instance ID of the partner")
+@click.option("-s", "--settings", required=True, help="Comma separated list of all setting names to be excluded")
+@click.pass_obj
+@pretty_print
+def upsert_policy_exception(controller, partner, expires=None, instance_id=None, policy_id=None, settings=None ):
```
makk94, line 389: rm the default param values

python/cli/prelude_cli/views/scm.py
```diff
@@ -355,6 +355,120 @@ def parse_from_partner_advisory(controller, partner, advisory_id):
...
+
+@policy_exceptions.command("delete")
+@click.argument(
+    "partner",
+    type=click.Choice(
+        [c.name for c in Control if c != Control.INVALID], case_sensitive=False
+    ),
+)
+@click.option("-i", "--instance_id", required=True, help="instance ID of the partner")
+@click.option("-p", "--policy_id", required=True, help="instance ID of the partner")
```
makk94, line 409: help text

python/cli/prelude_cli/views/scm.py
```diff
@@ -355,6 +355,120 @@ def parse_from_partner_advisory(controller, partner, advisory_id):
...
+)
+@click.option("-e", "--expires", help="Expiry Date (YYYY-MM-DD hh:mm:ss ([+-]hh:mm))", default=None, type=str)
+@click.option("-i", "--instance_id", required=True, help="instance ID of the partner")
+@click.option("-p", "--policy_id", required=True, help="instance ID of the partner")
+@click.option("-s", "--settings", required=True, help="Comma separated list of all setting names to be excluded")
+@click.pass_obj
+@pretty_print
+def upsert_policy_exception(controller, partner, expires=None, instance_id=None, policy_id=None, settings=None ):
+    """update policy exception"""
+    with Spinner(description=f"Updating/Updating Policy exceptions"):
```
makk94, line 391: rm one of the "updating"s?

makk94: oh was one of these meant to be "Creating" ? 
i've never seen `policy_exceptions.add_command(upsert_policy_exception, name="create")` but I assume that redirect `create` cmd to this function too?

garjarbinks: yes it tells click to reuse the function for create vs duplicating code - because the SDK is `put_policy_exception` for both with the same params

python/cli/prelude_cli/views/scm.py
```diff
@@ -355,6 +355,120 @@ def parse_from_partner_advisory(controller, partner, advisory_id):
             partner=Control[partner], advisory_id=advisory_id
         )
 
+@scm.group("exceptions")
```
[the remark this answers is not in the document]
garjarbinks: Done

python/cli/prelude_cli/views/scm.py
makk94: still need to update `Updating/Updating` ?

python/cli/prelude_cli/views/scm.py
```diff
@@ -355,6 +355,129 @@ def parse_from_partner_advisory(controller, partner, advisory_id):
...
+        [c.name for c in Control if c != Control.INVALID], case_sensitive=False
+    ),
+)
+@click.option("-e", "--expires", help="Expiry Date (YYYY-MM-DD hh:mm:ss ([+-]hh:mm))", default=None, type=str)
+@click.option("-i", "--instance_id", required=True, help="instance ID of the partner")
+@click.option("-p", "--policy_id", required=True, help="ID of the policy to update")
+@click.option("-s", "--settings", required=True, help="Comma separated list of all setting names to be excluded")
+@click.pass_obj
+@pretty_print
+def update_policy_exception(controller, partner, expires=None, instance_id=None, policy_id=None, settings=None ):
```
makk94, line 379: rm the `=None` on all the function params (the option decorator will define or default the default)

makk94: also have an extra space before your final parenthesis

python/cli/prelude_cli/views/scm.py
```diff
@@ -355,6 +355,129 @@ def parse_from_partner_advisory(controller, partner, advisory_id):
...
+        [c.name for c in Control if c != Control.INVALID], case_sensitive=False
+    ),
+)
+@click.option("-e", "--expires", help="Expiry Date (YYYY-MM-DD hh:mm:ss ([+-]hh:mm))", default=None, type=str)
+@click.option("-i", "--instance_id", required=True, help="instance ID of the partner")
+@click.option("-p", "--policy_id", required=True, help="ID of the policy to create")
+@click.option("-s", "--settings", required=True, help="Comma separated list of all setting names to be excluded")
+@click.pass_obj
+@pretty_print
+def create_policy_exception(controller, partner, expires, instance_id, policy_id, settings ):
```
makk94, line 403: extra space still on the rest of these :( 

we use the Black code formatter if you have it

PR #1225, opened 2025-07-22
Det 3227: cisco meraki

tangtan2:


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -483,6 +499,7 @@ class PolicyType(Enum, metaclass=MissingItem):
     DEVICE_COMPLIANCE = 11
     IDENTITY_MFA = 12
     HOST_FIREWALL = 13
+    NETWORK_FIREWALL = 16
```
makk94, line 502: idk if you're leaving room for something, but this PR is probably next to go in, so I'd jsut squish up the numbers

makk94 (approved):

PR #1237, opened 2025-08-11
det-3221: remove NO_HARDWARE, add CHIP policy type

makk94:


python/sdk/prelude_sdk/models/codes.py
```diff
@@ -417,7 +417,6 @@ class PartnerEvents(Enum, metaclass=MissingItem):
     NO_SERVER_MANAGER = 14
     MISSING_HOST_FIREWALL_POLICY = 16
```
packplusplus, line 418: whats up with the jump?

packplusplus (approved):
Looks right. My guess is when we kill a control number we never use it again, so I may be answering my own question.

PR #1304, opened 2026-04-13
det-4098: update VST tests to account for `updated`

stephanwampouille:


python/sdk/tests/test_build.py
```diff
@@ -136,10 +137,15 @@ def test_update_test(self, unwrap):
         pytest.expected_test["expected"]["crowdstrike"] = EDRResponse.PREVENT.value
         pytest.expected_test["name"] = updated_name
         pytest.expected_test["technique"] = "T1234.001"
+        pytest.expected_test["updated"] = res["updated"]
 
         diffs = check_dict_items(pytest.expected_test, res)
         assert not diffs, json.dumps(diffs, indent=2)
 
+        assert parse(res["updated"]).replace(tzinfo=timezone.utc) <= datetime.now(
```
tangtan2: i dont know if this assertion is super useful, wouldn't it be btter to assert that res['updated'] > pytest.expected_test['updated']? (obvi before you do the assignment on line 140)

stephanwampouille: Yeah I was doing something similar to how we check for tombstoned dates but I can do this!

tangtan2 (approved):

PR #1309, opened 2026-05-04
add Agent package with privilege rerouting and input handling

garjarbinks:
## Summary

- Introduces `go/tests/agent` as a coordination layer above `Endpoint`, adding `RunAsPrivilege`, `PrivilegeLabel`, `VstId`, and `ReadInputs` in support of `advanced_probe`
- Adds `HostNotVulnerable` alias in `endpoint.go` to correct the existing typo `HostNotVulnerabile` while keeping the old constant for backwards compatibility

## VSTv1 Behaviours (backwards compat)
The effect if run in non-advanced probe (<v3)
- `RequitePrivilege` - INSUFFICIENT PRIVILAGES
- `PrivilegeLabel` - empty string
- `VstId` - UUID of test
- `ReadInputs` - 5s delay then continue

## Test plan

- [x] `go test -count=1 ./` passes without root
- [x] `sudo go test -count=1 ./` passes `TestRunAsPrivilegePrivileged` as root
- [x] `TestRunAsPrivilegeExits` asserts exit code 109 via subprocess pattern

```
% go test -count=1 -v ./ 2>&1 | grep -E "^(---|PASS|FAIL|ok|SKIP)"

--- PASS: TestParseInputVariables (0.00s)
--- PASS: TestVstId (0.00s)
--- PASS: TestPrivilegeLabelNoMatch (0.00s)
--- PASS: TestPrivilegeLabelMatch (0.00s)
--- PASS: TestReadInputsEmptyPipe (0.00s)
--- PASS: TestReadInputsVariablesWrapper (0.00s)
--- PASS: TestReadInputsFlatMap (0.00s)
--- PASS: TestRunAsPrivilegeUnprivileged (0.00s)
--- SKIP: TestRunAsPrivilegePrivileged (0.00s)
--- PASS: TestRunAsPrivilegeExits (0.01s)
PASS
ok      github.com/preludeorg/libraries/go/tests/agent  0.715s
```

```
% sudo go test -count=1 -v -run TestRunAsPrivilegePrivileged ./ 2>&1 | grep -E "^(---|PASS|FAIL|ok|SKIP)"

--- PASS: TestRunAsPrivilegePrivileged (0.00s)
PASS
ok      github.com/preludeorg/libraries/go/tests/agent  0.309s
```


tangtan2 (approved):
i think it looks fine but im not a go expert so @packplusplus if you could take a look that would be great

go/tests/agent/agent.go
```diff
@@ -0,0 +1,167 @@
+// Package Agent provides utility functions for Prelude Detect agent control,
+// primarily for v3 agents and above with safe defaults for previous versions.
+package Agent
```
packplusplus: Ignoring what our existing endpoint library does, I think go standard is lower case for package names. https://go.dev/blog/package-names

go/tests/agent/agent.go
```diff
@@ -0,0 +1,167 @@
...
+		if n := numParts[0]; n < len(parts) {
+			parts = parts[:n]
+		}
+		return strings.Join(parts, ".")
+	}
+	return ver
+}
+
+// VstId derives the VST ID from the executable name, stripping any suffix after an underscore and the file extension.
+func VstId() string {
```
packplusplus: If GetPreludeVersion uses get a get prefix, this should too. 

Does it make sense to do it this way or take the test name from an env var as the source of truth? 

Also, initialism, should be VstID:  https://go.dev/wiki/CodeReviewComments#initialism

go/tests/agent/agent.go
```diff
@@ -0,0 +1,167 @@
...
+    label, found := strings.CutPrefix(filepath.Base(cwd), "PreludeHelper_")
+    if !found {
+        return ""
+    }
+    return strings.ToLower(label)
+}
+
+// ReadInputs reads JSON from stdin with a 5s timeout.
+// Expects {"Variables": {"key": "val"}} or a flat {"key": "val"} map.
+// on Pre 3.x systems will simply delay 5s then continue.
```
packplusplus: I find this comment confusing, or wrong. If someone spawns this and doesn't pass it a fh, it's going to get /dev/null for stdin and not have a delay. if it does get spawned with a pipe, it will wait 5 seconds for input and keep going, but why wait even 5 seconds? the agent start the program and is waiting on a call to a third party to feed the data to the tests? then 5 seconds is too short.

go/tests/agent/agent.go
```diff
@@ -0,0 +1,167 @@
...
+		Endpoint.Say("Error reading input: %v", err)
+		return map[string]string{}
+	case <-time.After(timeout):
+		Endpoint.Say("No input received within timeout, continuing...")
+		return map[string]string{}
+	}
+}
+
+// RunAsPrivilege checks if the current process meets the specified privilege level ("privileged", "unprivileged", or a custom label).
+func RunAsPrivilege(level string) {
```
packplusplus: this is a misleading function name, should be something like validate privs, not run as.

go/tests/agent/agent.go
```diff
@@ -0,0 +1,167 @@
...
+		return value
+	}
+	return defaultValue
+}
+
+// GetPreludeVersion retrieves the PRELUDE_VERSION environment variable, defaulting to "0.0.0.0" if unset or empty.
+// An optional numParts argument truncates the result to that many components (e.g. numParts=2 returns "1.2" from "1.2.3.4").
+// If omitted, zero, or greater than the number of components, the full version string is returned.
+func GetPreludeVersion(numParts ...int) string {
+	ver := GetPreludeEnvironmentVariable("VERSION", "0.0.0.0")
```
packplusplus: We use sem ver for all our other products, where did this quad number thing come from?

go/tests/agent/agent.go
```diff
@@ -0,0 +1,167 @@
...
+	case <-time.After(timeout):
+		Endpoint.Say("No input received within timeout, continuing...")
+		return map[string]string{}
+	}
+}
+
+// RunAsPrivilege checks if the current process meets the specified privilege level ("privileged", "unprivileged", or a custom label).
+func RunAsPrivilege(level string) {
+    var satisfied bool
+	var lowerlevel = strings.ToLower(level)
```
packplusplus: should be mixed caps, lowerLevel

Effective Go — Names: https://go.dev/doc/effective_go#names — official, covers package names, getters, interface names, and mixedCaps. .

go/tests/agent/agent.go
```diff
@@ -0,0 +1,167 @@
...
+    }
+    if !satisfied {
+        fmt.Fprintf(os.Stderr, "PRIVILEGE_REQUESTED:%s", lowerlevel)
+        Endpoint.Stop(Endpoint.InsufficientPrivileges)  // exit 109
+    }
+}
+
+// TolerateImpactOrExit stops execution with InsufficientPrivileges if the
+// given impact level exceeds PRELUDE_TOLERANCE (defaults to 0 if unset or invalid).
+func TolerateImpactOrExit(impact int) {
```
packplusplus: maybe similar to validate privs

go/tests/endpoint/endpoint.go
```diff
@@ -36,6 +36,7 @@ const (
 	FileQuarantinedOnExtraction int = 105
 	NetworkConnectionBlocked    int = 106
 	HostNotVulnerabile          int = 107
+	HostNotVulnerable          int = 107
```
packplusplus, line 40: mark deprecated, and `// Deprecated: use HostNotVulnerable instead.`

spacing

go/tests/agent/agent.go
```diff
@@ -0,0 +1,167 @@
...
+    name := filepath.Base(bin)
+    name = strings.TrimSuffix(name, filepath.Ext(name)) // strip .exe on Windows if present
+    if idx := strings.Index(name, "_"); idx != -1 {
+        return name[:idx]
+    }
+    return name
+}
+
+// PrivilegeLabel derives the privilege label from the current working directory name, expecting a format like "PreludeHelper_Label".
+func PrivilegeLabel() string {
```
packplusplus, line 81: If GetPreludeVersion uses get a get prefix, this should too.

go/tests/agent/agent.go
```diff
@@ -0,0 +1,167 @@
...
+    switch lowerlevel {
+		case "privileged":
+			satisfied = Endpoint.CheckAdmin()
+		case "unprivileged":
+			satisfied = !Endpoint.CheckAdmin()
+		default:
+			satisfied = lowerlevel == PrivilegeLabel()
+    }
+    if !satisfied {
+        fmt.Fprintf(os.Stderr, "PRIVILEGE_REQUESTED:%s", lowerlevel)
```
packplusplus: other errors put a space after the colon and a more descriptive error.

packplusplus (changes requested):
You need to run gofmt and golint on this, it's got a mix of tabs and spaces and non-standard line breaks. Consider a more strict linter, like revive or whatever and a pre-commit hook to execute them on commits to make sure we don't drift. 

Endpoint likely has similar issues, but we don't have to boil the ocean, just make sure new libraries are good and clean. 

I'd also maybe consider re-orging the file so privates and public functions are less intermingled, but that's a personal preference, and you could probably pull up go docs suggesting some other pattern to use.

go/tests/agent/agent.go
garjarbinks: TY! for consistency with others, Ive removed the Get prefixes

go/tests/agent/agent.go
garjarbinks: 100% correct.  there is no delay for pre v3.  The 5s though I believe to be sufficient time for the helper to feed the inputs via stdin to the VST.

go/tests/agent/agent.go
garjarbinks: RequirePrivielge

go/tests/agent/agent.go
garjarbinks: Renamed to TolerateImpact — the OrExit suffix is redundant once the exit behaviour is implied by the function's role as a precondition guard, same as RequirePrivilege

go/tests/agent/agent.go
garjarbinks: Fixed — changed default to "0.0.0" (semver). Will coordinate advanced probe too.,

go/tests/agent/agent.go
garjarbinks: removed Get prefixes per Go convention

packplusplus (approved):
bueno!

bfuzzy1 (approved):

